PR #4, opened 2022-01-24
1) YALB-71: Atoms: Dividers

ModulesUnraveled:
### Description of work
- Adds the initial Dividers component

### Functional testing steps:
- [x] View the [Divider story](http://localhost:6006/?path=/story/atoms-divider--spacing) and verify acceptance criteria are met
- [x] Choose settings in the Controls section of Storybook to see how the dividers are affected

### DIFF
--- a/components/00-base/functions/map.scss
+++ b/components/00-base/functions/map.scss
@@ -1,0 +1,12 @@
+@use 'sass:map';
+
+/// Map deep get
+/// @param {Map} $map - Map
+/// @param {Arglist} $keys - Key chain
+/// @return {*} - Desired value
+@function deep-get($map, $keys...) {
+  @each $key in $keys {
+    $map: map.get($map, $key);
+  }
+  @return $map;
+}
--- a/components/01-atoms/divider/_divider.scss
+++ b/components/01-atoms/divider/_divider.scss
@@ -18,7 +18,7 @@ $positions: left, center, right;
 }
 
 .divider {
-  background: var(--color-yale-black-gray);
+  background: var(--color-divider);
   width: 100%;
   height: calc(0.5rem / 16);
 
--- a/components/01-atoms/divider/_divider.scss
+++ b/components/01-atoms/divider/_divider.scss
@@ -3,19 +3,14 @@
 
 $widths: map.deep-get(tokens.$tokens, layout, width);
 $thicknesses: map.deep-get(tokens.$tokens, size, thickness);
-$positions: left, center, right;
+$positions: map.deep-get(tokens.$tokens, layout, flex-position);
 
 .divider__wrapper {
   display: flex;
-  justify-content: center;
 
-  @each $position in $positions {
+  @each $position, $value in $positions {
     &[data-divider-position='#{$position}'] {
-      @if $position == left {
-        justify-content: flex-start;
-      } @else if $position == right {
-        justify-content: flex-end;
-      }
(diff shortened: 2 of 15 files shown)


components/_settings/_config.scss
```diff
@@ -0,0 +1,4 @@
+:root {
+  --color-accent: var(--color-theme-accent, var(--color-blue-medium));
+  --color-divider: var(--color-gray-500);
```
nJim, line 3: I see this declared here, but I wasn't sure how it gets into the components. When running storybook I don't have `--color-divider` defined.

ModulesUnraveled: Weird, I do see it defined. If you want, I can look at it with you, just to make sure. Otherwise, we'll continue assuming my environment is correct, and that it's there. If it's not somehow, we'll notice it pretty quick.

components/01-atoms/divider/_divider.scss
```diff
@@ -0,0 +1,36 @@
+$widths: 100%, 75%, 50%, 25%;
```
callinmullaney, line 1: Maybe you did this in a recent PR that I havent got to yet but it seems that these values would be better defined as tokens. Widths could also relate to grids, thickness may relate to borders, positions could be used for alignments in other components. You could probably re-assign those token values in this file to be something that is divider name-spaced in this file that you then loop through.

Ignore comment if I just haven't got to that change yet. Otherwise, I dig the logic in this file and re-assigning left/right to flex specific values makes sense.

components/01-atoms/divider/divider.twig
```diff
@@ -0,0 +1,9 @@
+{% set divider__base_class = 'divider' %}
```
callinmullaney, line 1: I think there should be comments here for the available variables like we've done for projects in the past. Because this is the first component it might be good to establish this practice now instead of retro-actively updating things.

components/_settings/_config.scss
callinmullaney: I noticed this was updated in a later PR which I used to generate a throw-away theme on my local. Can confirm Brian's comments here.

callinmullaney (changes requested):
@ModulesUnraveled See some comments I left. I'm cool with approving if you want to follow up in another branch. I know you've been chaining these so going back and updating isn't always the most efficient.

components/01-atoms/divider/divider.twig
ModulesUnraveled: Thanks for reminding me about this! Done.

components/01-atoms/divider/_divider.scss
ModulesUnraveled: I hadn't done this yet, and I was able to figure it out. Should be there in the latest commits.

If you're curious, here's the associated tokens repo PR: https://github.com/yalesites-org/tokens/pull/12

components/00-base/functions/map.scss
```diff
@@ -0,0 +1,12 @@
+@use 'sass:map';
+
+/// Map deep get
```
callinmullaney, line 3: Good call adding the deep-get function. This is a lot easier to read/use compared to using nested `map-get`'s

callinmullaney (approved):
@ModulesUnraveled This looks great on my end. Thanks for implementing my feedback from earlier. :shipit:

PR #6, opened 2022-02-01
4) YALB-65: Base: Breakpoints

ModulesUnraveled:
### Description of work
- Defines the breakpoints for the site

### Functional testing steps:
- [ ] View the breakpoints story... that's really it


.storybook/preview.js
```diff
@@ -6,7 +6,7 @@ import { setupTwig } from './setupTwig';
 // GLOBAL CSS
 import '../components/style.scss';
 import '../components/00-base/colors/cl-colors.scss';
-import '../components/00-base/spacing/cl-spacing.scss';
+import '../components/00-base/base/cl-spacing.scss';
```
callinmullaney, line 9: Renaming `00-base` to `00-tokens` in the more recent PR is a good call as this seems a bit redundant. That being said, what else are you thinking would have a home in `00-tokens/base/`? I'm thinking this should be renamed to `00-tokens/breakpoints/`.

callinmullaney (changes requested):
See code comment and if you believe this warrants a change in this PR.

Something I think you should consider for this breakpoints, colors, spacing, and future tokens/components that display the usage. For these stories you're featuring the usage by displaying the planned scss variable or custom property. It would be a nice-to-have feature for new devs to quickly view and copy these use cases. I did this in the old clr function PR here: https://github.com/emulsify-ds/emulsify-drupal/pull/191/files#diff-74410c1b391b7fbd701c702643444e956e5aaa944e1e559f2d809e9efcdee629R11 

Example: If a dev is building out a new component and has storybook open in a window it would be convenient to click on the breakpoint they want to implement and it automatically copies `@include breakpoint($break-m) {}` to their clip board. They could create this as a auto-complete code snippet in their IDE but I figure this copy method would be less effort on their part while still making the efficiency available for all contributors.

callinmullaney (changes requested):
See code comment and if you believe this warrants a change in this PR.

Something I think you should consider for this breakpoints, colors, spacing, and future tokens/components that display the usage. For these stories you're featuring the usage by displaying the planned scss variable or custom property. It would be a nice-to-have feature for new devs to quickly view and copy these use cases. I did this in the old clr function PR here: https://github.com/emulsify-ds/emulsify-drupal/pull/191/files#diff-74410c1b391b7fbd701c702643444e956e5aaa944e1e559f2d809e9efcdee629R11 

Example: If a dev is building out a new component and has storybook open in a window it would be convenient to click on the breakpoint they want to implement and it automatically copies `@include breakpoint($break-m) {}` to their clip board. They could create this as a auto-complete code snippet in their IDE but I figure this copy method would be less effort on their part while still making the efficiency available for all contributors.

.storybook/preview.js
ModulesUnraveled: I'll make this change in my current branch (Color Pairings) because other things have changed. That way I don't have to do the merge-conflict dance.

To answer your question, though, "I don't know." In the past, we've put things like breakpoints, mixins, and base styles like box-sizing in the `base` folder. Most of that won't be needed on this project since they'll come from the tokens in some form. For now, I've started putting the "base" styles (like the stuff below) in the main `style.scss` file along with the imports. We'll see if that needs to move (i.e. if it gets too big) later.

```
:root {
  --color-text: var(--color-basics-black);
}

html {
  box-sizing: border-box;
}

*,
*::before,
*::after {
  box-sizing: inherit;
}

body {
  font: var(--font-style-body-default);
}
```

kara-franco (changes requested):
Breakpoint information: put into a data table <table> to give more semantic meaning for screen reader users. 

Since we will be embedding this info into a public website, I want to be sure that a screen reader user can easily navigate the info and associate table headers with data. 

[https://www.w3.org/WAI/tutorials/tables/one-header/](https://www.w3.org/WAI/tutorials/tables/one-header/)

ModulesUnraveled:
@kara-franco Thanks, take a look at it now (and also the spacing story) and let me know if the new markup works better for you

.storybook/preview.js
callinmullaney: For those styles specifically I see why you would add them to `style.scss` and that makes sense to a certain degree I think. But given this project is more of a platform I could see these styles living in `00-tokens/00-reset/` or `00-tokens/00-foundation/` or something like that.

kara-franco (approved):
Excellent, thank you so much!

callinmullaney (approved):

PR #7, opened 2022-02-01
5) YALB-67: Effects

ModulesUnraveled:
### Description of work
- Creates Shadows, Radii, and Borders

### Functional testing steps:
- [ ] View the [Shadows](http://localhost:6006/?path=/story/tokens-effects--shadows) and other Effects. Verify everything looks as expected.


miketullo95:
I'm curious what other people think about drop-shadow level 1. It seems very slight and wondering if people might default pick 1 and notice no difference. @ModulesUnraveled  @kara-franco

callinmullaney (changes requested):
@ModulesUnraveled Looking pretty good here. Only see some minor inconsistencies:
1. Examples for Shadow blocks are blue, Radii are grey, and borders are black. Unless this is intentional it seems these should be the same color... 
2. Borders is missing an `<h2>` heading.

ModulesUnraveled:
The colors are the same that are defined in Figma, so I'm going to leave them so that they're consistent. And the "docs" at the top are just some notes that were in the designs, and I figured I'd bring over here for now (they'll eventually live in the documentation site). I don't think there needs to be an `<h2>` in this case.

callinmullaney (approved):

ModulesUnraveled:
@miketullo95 I wondered the same thing, but that's a question that should go to design. This is just a 1:1 representation of what's in Figma. If it should be updated, there should be a design cycle to address it, and then we can update the code to match.

PR #10, opened 2022-02-03
YALB-70: Atoms: Links

ModulesUnraveled:
### Description of work
- Creates the initial `link` component and styles
- Implements links on color pairings example
- Creates a custom PR template for this repo

### Testing Setup
- [x] Checkout this branch
- [x] Run `npm install`
- [x] Run `npm run develop`
- [x] View the [Links Story](http://localhost:6006/?path=/story/atoms-links--links)
- [x] View the [Color Pairings Story](http://localhost:6006/?path=/story/tokens-colors--color-pairings)

### Functional Review Steps
- [x] Verify the links are on the page, and function as expected

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/file/l84dYRVCvA5DbbCkG0yWke/UI-Kit-%5BYaleSites%5D)

### Accessibility Review

## Notes:
- The current implementation is missing icons since those are still in review/being decided


components/01-atoms/links/_link.scss
```diff
@@ -0,0 +1,59 @@
+@mixin animation-link($property) {
+  transition-property: #{$property};
+  transition-duration: 0.2s;
+  transition-timing-function: ease-in-out;
+}
+
+a {
```
ModulesUnraveled, line 7: @callinmullaney I'm styling the `<a>` tag directly for now. I expect I'll have to refactor it and scope it to `.text-field a` or something like that eventually, but until I hit an issue and/or create the general `text` component that would implement that, I don't know how that should be handled yet.

ahitson:
Color pairings: make the headlines 800 and the body copy 700 when they are on dark backgrounds. All links should be in the medium/500 weight.

Links: change to medium/500 weight.

ModulesUnraveled:
@ahitson I made those updates

kara-franco (changes requested):
Looks great! The only color pairing that needs to be addressed is the text hover color on gray-700. The blue should target 4.5:1 color contrast for normal text. 

Screen Shot 2022-02-08 at 11 55 21 AM

kara-franco:
Excellent, fixes looks great.

components/01-atoms/links/link.twig
```diff
@@ -0,0 +1,23 @@
...
+ # - link_attributes - array of attribute,value pairs
+ #
+ # Available blocks:
+ # - link_content - used to replace the content of the link
+ #     Example: to insert the image component
+ #}
+
+{% set link_base_class = link_base_class|default('link') %}
+
+<a {{ bem(link_base_class, link_modifiers, link_blockname) }}
```
callinmullaney, line 14: Should we have a default `aria-label` attribute at this level or is there a way we can help enforce this where this component is used. Looking at Yale's accessibility rules it seems like `aria-label` or `aria-labelledby` or `aria-describedby` must be present. https://usability.yale.edu/web-accessibility/articles/links

Also, instead of using a for loop here can we take advantage of the `add_attributes` function somehow?

```
{% set link_attributes = link_attributes|default({
  'class':  bem(link_base_class, link_modifiers, link_blockname),
}) %}

<a add_attributes(link_attributes)>
```

callinmullaney (changes requested):
@ModulesUnraveled couple comments/questions to address.

components/01-atoms/links/_link.scss
```diff
@@ -0,0 +1,63 @@
...
+  --color-link-hover: var(--color-text);
+
+  @include animation-link('color');
+
+  position: relative;
+  color: var(--color-link);
+  text-decoration: none;
+  font-weight: var(--font-weights-mallory-500);
+
+  &:hover {
```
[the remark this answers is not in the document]
ModulesUnraveled: I actually thought the same thing, and had them both changing together, but was asked to keep the link color unchanged on focus.

components/01-atoms/links/link.twig
```diff
@@ -0,0 +1,23 @@
...
+ # - link_url - the url this link should poing to
+ # - link_attributes - array of attribute,value pairs
+ #
+ # Available blocks:
+ # - link_content - used to replace the content of the link
+ #     Example: to insert the image component
+ #}
+
+{% set link_base_class = link_base_class|default('link') %}
+
```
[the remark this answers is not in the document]
ModulesUnraveled: Yes, there's actually a note in the PR details that says the icons are still undecided (whether we're going to use Font Awesome, or something else) so that'll come in a later PR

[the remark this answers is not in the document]
callinmullaney: ah! missed the note. All good here then :+1:

components/01-atoms/links/link.twig
ModulesUnraveled: I don't think we need an aria-label right now. That would come in handy if it were on a card, and the link text was "Read more" then we used the card heading as the aria label so that it read out "Read more Awesome card heading" or something like that. A link by itself doesn't need an aria-label unless there is no text (like an icon/button) or if the text isn't helpful on its own (like "Read More")

And YES! I will use the add_attributes thing. I copied this from Compound as a starting point, and I didn't even consider updating it. On it now.

ModulesUnraveled: @callinmullaney done

components/01-atoms/links/link.twig
```diff
@@ -0,0 +1,25 @@
...
+{# If link_attributes is not defined, set it to an empty object by default #}
+{% set link_attributes = link_attributes|default({}) %}
+
+{% set link_attributes = link_attributes|merge({
+  class: bem(link_base_class, link_modifiers, link_blockname)
+}) %}
+
```
callinmullaney, line 20: I think this would be cleaner but correct me if I'm wrong 
```
{% set link_attributes = link_attributes|merge({
  class: bem(link_base_class, link_modifiers, link_blockname),
  href: link_url
}) %}
```

callinmullaney (changes requested):

ahitson (approved):

components/01-atoms/links/link.twig
ModulesUnraveled: Ooh, yeah! I like that

ModulesUnraveled: Done. Totes cleaner!

callinmullaney (approved):
@ModulesUnraveled Thanks for making those changes. :shipit:

PR #11, opened 2022-02-06
refactor: updated thickness names to match tokens

ModulesUnraveled:
### Description of work
- This is just a cleanup ticket to align the thickness names in the component library with the token names
- 
### Testing Setup
- [ ] Checkout this branch
- [ ] Run `npm ci`
- [ ] Run `npm run develop`
- [ ] Verify the [Border Thickness](http://localhost:6006/?path=/story/tokens-effects--borders) and [Divider](http://localhost:6006/?path=/story/atoms-divider--dividers&args=position:!undefined;dividerColor:!undefined) stories both use the scale "hairline, 1, 2, 4, 6, 8"

### DIFF
--- a/components/01-atoms/divider/divider.stories.js
+++ b/components/01-atoms/divider/divider.stories.js
@@ -9,9 +9,9 @@ export default {
       defaultValue: 'center',
     },
     thickness: {
-      options: [1, 2, 3, 4, 5, 6],
+      options: ['hairline', 1, 2, 4, 6, 8],
       type: 'select',
-      defaultValue: 1,
+      defaultValue: 'hairline',
     },
     dividerColor: {
       options: ['gray-500', 'blue-yale', 'accent'],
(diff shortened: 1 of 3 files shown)


components/01-atoms/divider/divider.stories.js
```diff
@@ -22,12 +22,12 @@ export default {
 };
 
 export const Dividers = ({ position, thickness, dividerColor }) => `
+  ${dividerTwig({ divider__thickness: 'hairline' })}<br />
```
callinmullaney, line 25: At first I thought this was weird you were skipping 3 and 5. Then I realized after looking at the token definitions that this was a multiplier. 

Hairline does seem like the weird outlier here and I'm wonder if there is a better way to address this? Possible ideas:
- `thin` (easier to type)
- Refactor `hairline` to be 1 and adjust the others accordingly.
- `$size-thickness-0-5` given it is half the size of `$size-thickness-1`

callinmullaney (commented):
@ModulesUnraveled See comment. Not sure if this warrants any changes.

components/01-atoms/divider/divider.stories.js
callinmullaney: I'm not going to die on this hill though.

ModulesUnraveled: Yeah, I had the same concerns and talked with the designers about it. We tried to brainstorm some different ways, but couldn't come up with anything better. So we'll stick to this for now

PR #12, opened 2022-02-07
YALB-72: Atoms: Buttons

ModulesUnraveled:
### Description of work
- Creates the button atom
- Styles the button for all component themes

### Testing Setup
- [x] Checkout this branch
- [x] Run `npm ci`
- [x] Run `npm run develop`
- [ ] Navigate to the [Buttons Story](http://localhost:6006/?path=/story/atoms-button--button)
- [ ] Navigate to the [Color Pairings Story](http://localhost:6006/?path=/story/tokens-colors--color-pairings)

### Functional Review Steps
- [x] Verify everything looks correct, and works as expected

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/file/l84dYRVCvA5DbbCkG0yWke/UI-Kit-%5BYaleSites%5D)

### Accessibility Review


ahitson:
Buttons:
- Focus should be medium blue
- change text to medium/500 weight
- the buttons feel LARGE. How much can we reduce the padding and keep it accessible?

Color Pairings:
- Cap the first letter in the buttons to show style preference
- Add space between buttons like how a button group would show (32px)

ModulesUnraveled:
@ahitson I've updated the font-weight, focus color, and spacing locally. Let's talk about the padding and first letter thing.

Tl;dr:
- I'm using the closest padding token (16px) for now, but it's common to have different spacing values for `margin` things like the spacing _between_ things - and spacing values for `padding` which is the spacing inside things (like buttons and cards).
- I'm also curious if the first letter caps thing should be enforced by code, or just demoed as an example for _preference_, but not prevent people from using lowercase, or all caps, etc.

kara-franco (approved):
Updates look good. @ahitson The minimum button width/height we want is 44px on any given side, so we could definitely make these smaller, if preferred. [https://www.w3.org/WAI/WCAG21/Understanding/target-size.html](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html)

re: ARIA roles
If I wanted to create a disclosure button, would I add that prop like this? 
```
   ${button({
      button_content: 'Disclosure Button',
      button_radius: 'pill',
      aria_expanded: 'false',
    })}
```
EDIT! 
I just realized the focus indicator on the blue-yale background needs to be lightened to a lighter blue (to match the link focus indicator). Color contrast of focus indicators should be 3:1, currently the medium blue is at 2.3:1. 
Screenshot: 
Screen Shot 2022-02-08 at 11 36 23 AM

ahitson:
Only comment is to make the focus indicator the same as the link color.

ModulesUnraveled:
@ahitson I think all of your change requests are in! (The text colors will be fixed in the links PR)

@kara-franco We updated the focus outline to match the link color for each color pairing, so if you want to re-review those, that would be good.

Also, yes, that's the plan for disclosures (I'll use it for menus, etc.) Right now, it's just theory but when I actually implement it, I'll validate that everything works as expected. (And you're required to pass the `aria_controls` id so that it knows what it's expanding)

kara-franco:
Fix looks great!

ahitson:
Yay! (approved)

ahitson (approved):

components/01-atoms/buttons/_buttons.scss
```diff
@@ -0,0 +1,86 @@
+$button-radii: (
+  'soft': var(--button-radius-soft),
+  'pill': 100vw,
```
callinmullaney, line 3: Should this be converted to a token value instead? `$layout-width-full` ?

components/01-atoms/buttons/_buttons.scss
```diff
@@ -0,0 +1,86 @@
...
+);
+$button-style: (
```
callinmullaney, line 5: I don't see where these `$button-style` value are being used? Removing them had no affect that I could see. If these are needed I would think they should use the `$size-spacing-5` and `$size-spacing-7` or some re-assignment of these tokens.

components/01-atoms/buttons/_buttons.scss
```diff
@@ -0,0 +1,86 @@
...
+.button {
+  --color-button-outline: var(--color-link);
+
+  border: var(--border-thickness-button, var(--border-thickness-2)) solid
+    var(--color-button-border);
+  border-radius: var(--border-radius-button, var(--radius-0));
+  background: var(--color-button-bg);
+  color: var(--color-button-text);
+  font-weight: var(--font-weights-mallory-500);
+  padding: var(--size-spacing-4) var(--size-spacing-7);
```
callinmullaney, line 24: add `cursor:pointer;` maybes???

callinmullaney (changes requested):

callinmullaney:
@ModulesUnraveled some feedback and a merge conflict for you to address :sandwich:

components/01-atoms/buttons/_buttons.scss
ModulesUnraveled: I considered this, but the "pill" (`100vw`) radius is only used by buttons, and should not be accessible to anything else. If I add it to the `radius` array, then it shows up as a border-radius option for things like cards too, and I don't want that. If it does come up again, maybe I'll see about moving it out of this file, but for now, I think this is the right place for the definition.

(Also, `layout-width-full` is 100% which is very different from 100vw)

oh... I actually just realized I should be using 100vmax... I'll make that change.

components/01-atoms/buttons/_buttons.scss
ModulesUnraveled: Nope. I think I was going to, but when you look lower, there are enough fiddly differences that I couldn't do a simple @each like the radii and weights.

I've deleted it.

components/01-atoms/buttons/_buttons.scss
ModulesUnraveled: Yeah, I know we do that a lot for buttons, but I've also heard that the pointer should be reserved for links (`<a>` tags).

@kara-franco do you have any info on whether there's an accessibility reason to use or not use the pointer cursor for buttons?

Here's a reference, but it doesn't give advice on when to use one or another: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

components/01-atoms/buttons/_buttons.scss
callinmullaney: Gotcha. All makes sense here and I agree. I just saw a static value being set and wanted to confirm this was the ideal setup.

components/01-atoms/buttons/_buttons.scss
callinmullaney: I've heard that as well but I've also been under the impression that any element that can be clicked by a user should use a pointer. I'm going to approve this PR but lets wait on merging until @kara-franco chimes in.

callinmullaney (approved):

components/01-atoms/buttons/_buttons.scss
ModulesUnraveled: @kara-franco Thanks for the references! I've shared them internally as well.

PR #13, opened 2022-02-09
YALB-83: Pull-Quote

ModulesUnraveled:
### Description of work
- Creates the Pull Quote component
- Implements the update to the drop shadow level 1

### Testing Setup

- [ ] Navigate to the [Pull Quote Story](https://deploy-preview-13--dev-component-library-twig.netlify.app/?path=/story/molecules-pull-quote--pull-quote)

### Functional Review Steps
- [ ] Verify the quotes look and function as expected

### Design Review
- [ ] Verify the pull quote designs match the [Figma Designs](https://www.figma.com/file/l84dYRVCvA5DbbCkG0yWke/UI-Kit-%5BYaleSites%5D)
- [ ] Verify the drop shadow level 1 matches the new figma design

### Accessibility Review


ahitson:
Realizing gray 800 is the wrong color. It should be 222222 (or RGB 34 34 34).

The quote icon is off. Please check against the design for the more "rounded w/ ball" type quote.

miketullo95:
Morning! Everything looks great! I did notice one small difference in the quote style for the parentheses. 

Screen Shot 2022-02-09 at 11 01 18 AM
Screen Shot 2022-02-09 at 11 01 14 AM

ahitson (approved):

kara-franco (approved):
Looks great! 

Will we allow the use of the cite attribute? I really have no idea how common is to use (I've never seen it used in the wild). I think in most of Yale's use cases, we are quoting students/faculty/authors more than actual external sources.

ModulesUnraveled:
@kara-franco We talked about that and decided that the purpose of this component is to quote an individual/organization/etc. and not to include the cite attribute. If we find we need that functionality that would be a different component.

One of the questions we had was about formatting, since we figured different groups may have different citation requirements (MLA etc.) and we didn't want to solve that now, since it would be the edge-case use of the component.

components/02-molecules/pull-quote/pull-quote.twig
```diff
@@ -0,0 +1,20 @@
...
+ #
+ # Available Variables:
+ # - pull_quote__quote (required)
+ # - pull_quote__attribution
+ #}
+
+{% set pull_quote__base_class = 'pull-quote' %}
+{% set pull_quote__style = pull_quote__style|default('bar-left') %}
+
+<figure {{ bem(pull_quote__base_class) }} data-pull-quote-style="{{ pull_quote__style }}">
```
callinmullaney, line 13: Similar feedback I had for the links. We should use the `add_attributes(pull_quote_attributes)` implementation in case this needs to be extended during implementation. https://github.com/yalesites-org/component-library-twig/pull/10/files#diff-e4131c9c0a0500830af2795a53a11107872df87c88b69e073bf5c396236206fcR17

callinmullaney (changes requested):
@ModulesUnraveled Nice! Added some small feedback to the twig file.

components/02-molecules/pull-quote/pull-quote.twig
ModulesUnraveled: Good idea. I've actually not really used the add_attributes function much in the past. But I think I like the pattern for the things we've done so far on this project. I'll try to remember to do it by default in the future!

callinmullaney (approved):
@ModulesUnraveled :shipit:

PR #17, opened 2022-02-09
YALB-107: Layouts

ModulesUnraveled:
### Description of work
- Creates a "placeholder" component for the component library
- Creates the initial "Full Width" page layout (It's not really exciting right now, but will be soon!)

### Testing Link(s)
- [ ] Navigate to the [Placeholder Story](https://deploy-preview-17--dev-component-library-twig.netlify.app/?path=/story/page-layouts-utility--place-holder) preview
- [ ] Navigate to the [Full Width Story](https://deploy-preview-17--dev-component-library-twig.netlify.app/?path=/story/page-layouts-page-layouts--full-width) preview

### Functional Review Steps
- [ ] Verify the components show up :)

### Design Review
- [x] Nothing for now

### Accessibility Review
- [ ] Verify the proper elements and other markup are used (like `<main>` etc.)


components/04-page-layouts/_base.twig
```diff
@@ -0,0 +1,32 @@
+{% set layout__base_class = 'layout-container' %}
+{% set content__base_class = 'main' %}
+{% set sidebar__base_class = 'sidebar' %}
+{% set main__base_class = 'main-content' %}
+
+<div {{ bem(layout__base_class) }}>
+  {# Page Header #}
```
kara-franco, line 7: For the Header and Footer: can we use <header> and <footer> elements? These will provide more semantic meaning for screen reader users.

components/04-page-layouts/_base.twig
```diff
@@ -0,0 +1,32 @@
...
+  {% block page__header %}
+    {% include "@page-layouts/placeholder/placeholder.twig" with {
+      placeholder: 'Header',
+    } %}
+  {% endblock %}
+  {# System #}
+  {% block page_system %}{% endblock %}
+  <div {{ bem(content__base_class) }}>
+    <a id="main-content" tabindex="-1"></a>{# link is in html.html.twig #}
```
kara-franco, line 16: Needs a skip to main content in the header to jump here.

callinmullaney: @ModulesUnraveled second this feedback. I think you can migrate the id to the `<main>` instead of using a hidden `<a>` here. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main

callinmullaney (approved):
Looks good other than a comment that piggy-backs off of Kara's. Approving.

components/04-page-layouts/_base.twig
ModulesUnraveled: Yes, we will. I haven't actually created the header and footer components yet (those are coming next sprint.) The semantic markup you mentioned will be in those components.

This "layout" component is using Twig "block"s which are basically "buckets" where components can be inserted.

All of the `<section class="placeholder">` elements will be entirely replaced.

components/04-page-layouts/_base.twig
ModulesUnraveled: The skip link will actually be in a totally different file (a drupal templates called `html.html.twig`).

The code in this PR will be implemented in the `page.html.twig` Drupal template which will be inserted as the `{{ page }}` element in this example `html.html.twig`:
```
<!DOCTYPE html>
<html>
  <head>
    <title>{{ head_title|safe_join(' | ') }}</title>
  </head>
  <body>
    {# Keyboard navigation/accessibility link to main content section in page.html.twig. #}
    <a href="#main-content" class="visually-hidden focusable skip-link">
      {{ 'Skip to main content'|t }}
    </a>
    {{ page }}
  </body>
</html>
```

(I've simplified it for brevity, but things like google analytics tracking code, etc. will also e here.)

Not sure if that makes sense or not, but the point is the actual skip link will be in another file. This file just needs to have the target where the skip link takes the user to.

I do agree that the id doesn't need to be on an `<a>` tag, though, so I'll move the ID up to the content element.

ModulesUnraveled:
@callinmullaney @kara-franco I think I've addressed all of your feedback through either comments or changes. Let me know if there's anything else you notice

components/04-page-layouts/_base.twig
ModulesUnraveled: I started this PR this morning. Maybe it'll help understand things, in case I was just confusing up above. https://github.com/yalesites-org/component-library-twig/pull/18

components/04-page-layouts/_base.twig
kara-franco: I see, that makes sense! It's interesting to see how everything comes together.

kara-franco (approved):

callinmullaney (approved):
@ModulesUnraveled Code changes look good and follow-up comments make sense to me :shipit:

PR #18, opened 2022-02-10
YALB-98: Organisms: Site Header

ModulesUnraveled:
### Description of work
- Creates the Site Header and places it in the layout

### Testing Link(s)
- [ ] [View the Site Header Story](https://deploy-preview-18--dev-component-library-twig.netlify.app/?path=/story/organisms-site--header)
- [ ] [View the Site Header Examples Story](https://deploy-preview-18--dev-component-library-twig.netlify.app/?path=/story/organisms-site--header-examples)
- [ ] [View the Full Width Story](https://deploy-preview-18--dev-component-library-twig.netlify.app/?path=/story/page-layouts-page-layouts--full-width)

### Functional Review Steps
- [ ] Verify things all look and function as expected

### Design Review

### Accessibility Review


kara-franco:
Looks good! The placeholders in the utility nav and the main nav will be replaced with <nav> elements? And, if I remember correctly, the skip link will be added in as a separate task?

ahitson:
Please make the divider line under the utility nav to be grey 300. If it needs to pass accessibility, then whatever is the first step into accomplishing that is cool with me.

kara-franco:
> @ModulesUnraveled Markup and overall header organism look good from a technical standpoint. Couple of comments that may warrant a change in this PR but wanted to run them passed you first:
> 
> 1. Should the header have an optional `wrapper` inside `<header>` that imposes a `max-width`? It's hard to tell from the Figma designs if this is the case but as this PR stands now it will always be 100vw. Unless the plan is to set this max-width on the page wrapper itself.
> 2. The navigation order for keyboard users looks like it will be Yale Logo -> Utility Nav -> Site Logo -> Primary Nav. I'm wonder if the markup should be reorganized or tabindex attribute additions that allow this order to be Yale Logo -> Site Logo -> Primary Nav -> Utility Nav.

@callinmullaney I suggest having the tabbing order match the visual order on the page.

ModulesUnraveled:
@ahitson I think I've addressed all of your requests.

@miketullo95 This does not address the navigation. Right now I have it represented by a simple string (so that we could see how long lists of menu items will wrap etc.) but the actual navigation will come later when I get to the Primary Nav ticket. (Anything in a dashed border is just a placeholder for now, simply used to represent that something will be in that space eventually)
And thank you for pointing out the links, I've updated those so they should go to the appropriate places.

@callinmullaney I'm creating the wrapper styles like we talked about on zoom.
And I think I agree with Kara that the tab order should read left-to-right, then top-to-bottom like you're reading a book

@kara-franco Yes, the `<nav>` elements will come when I work on the menu components, and the skip link will be in another file all together

_... Phew!_

ahitson (approved):

miketullo95 (approved):
Approved UX/Front end

kara-franco (approved):

ModulesUnraveled:
@callinmullaney I added the wrapper we talked about. This should be ready for your review again!

components/00-tokens/functions/_map.scss
```diff
@@ -1,12 +1,18 @@
...
 /// @param {Map} $map - Map
 /// @param {Arglist} $keys - Key chain
 /// @return {*} - Desired value
 @function deep-get($map, $keys...) {
-  @each $key in $keys {
-    $map: map.get($map, $key);
+  $scope: $map;
+  $i: 1;
+  // stylelint-disable-next-line
+  @while (meta.type-of($scope) == map) and ($i <= list.length($keys)) {
```
callinmullaney, line 13: :clap:

callinmullaney (approved):
@ModulesUnraveled This is awesome! The inner styles should be really useful as you develop specific components in the future without the need to add repeat styles or mixins.

Also, great modification with the deep-map function and while loop. I took your idea here and played around with the old clr function PR I had. Seems like this function could be useful for a variety of scenario's that aren't token specific. 

:shipit:

PR #24, opened 2022-02-18
fix: defined button hover effects

ModulesUnraveled:
### Description of work
- Defines button hover effects

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/atoms-button--button) preview

### Functional Review Steps
- [ ] Verify the buttons each transition as expected

### Design Review

### Accessibility Review

### DIFF
--- a/components/01-atoms/buttons/_buttons.scss
+++ b/components/01-atoms/buttons/_buttons.scss
@@ -105,16 +105,22 @@ $button-outline-weights: (
   }
 
   &:active {
+    --color-button-bg-active: var(--color-gray-700);
+    --color-button-border-active: var(--color-gray-700);
+    --color-button-text-active: var(--color-basics-white);
+    --button-box-shadow-active: none;
+
     // Don't ever use !important!!!
-    // With that in mind, these are !important because the active styles below
+    // With that said, these are !important because the active styles below
     // should take precedence over any hover styles regardless of specificity.
     // We experimented with adding extra specificity to this selector to make it
     // "naturally" override the nested data attributes above... but that seemed
-    // more hacky than adding a few !importants
-    background-color: var(--color-gray-700) !important;
-    border-color: var(--color-gray-700);
-    color: var(--color-basics-white) !important;
-    box-shadow: none !important;
+    // more hacky than adding a few !importants. In this case we want the active
+    // state to take precedence over the hover state in any/all situations.
+    background-color: var(--color-button-bg-active) !important;
+    border-color: var(--color-button-border-active);
+    color: var(--color-button-text-active) !important;
+    box-shadow: var(--button-box-shadow-active) !important;
   }
 }
 
--- a/components/01-atoms/buttons/_buttons.scss
+++ b/components/01-atoms/buttons/_buttons.scss
@@ -33,6 +33,15 @@ $button-outline-weights: (
     outline-offset: var(--size-thickness-2);
   }
 
+  &:active {
+    // These are !important to override any hover styles defined by more
+    // specific selectors in the 'Hover styles' section below.
+    background-color: var(--color-gray-700) !important;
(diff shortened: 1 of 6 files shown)


ahitson:
- 200 does feel a bit fast on the Wipe, can we go 250? 
- Does the rise option have a drop shadow?
- Should click actions be here as well?
- Please add some space between each section so that the headings look like they are a part of what follows.

ModulesUnraveled:
@ahitson I updated the swipe duration, added the drop-shadow, and defined `active` (when the mouse is down) styles. Let me know if there's anything else!

miketullo95 (commented):
hello! after looking at the button wipe animation, I still feel like it's a little too fast and jarring. Could we experiment with the speed and see what 400 looks like (I think I'm looking at 250 right now).

components/01-atoms/buttons/_buttons.scss
```diff
@@ -29,6 +34,15 @@ $button-outline-weights: (
     outline-offset: var(--size-thickness-2);
   }
 
+  &:active {
+    // These are !important to override any hover styles defined by more
+    // specific selectors in the 'Hover styles' section below.
+    background-color: var(--color-gray-700) !important;
```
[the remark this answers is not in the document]
ModulesUnraveled: I updated the comments here according to our zoom conversation

ModulesUnraveled:
@miketullo95 @ahitson I've update the "wipe" version to use a 400ms transition duration. Let me know if that's better

components/01-atoms/buttons/_buttons.scss
```diff
@@ -64,19 +69,62 @@ $button-outline-weights: (
...
+      box-shadow: inset -6.5em 0 0 0 var(--color-button-bg);
+
+      &:hover {
+        box-shadow: none;
+      }
+    }
+  }
+
+  &:active {
+    // Don't ever use !important!!!
```
callinmullaney, line 108: 👍

components/01-atoms/links/_link.scss
```diff
@@ -1,6 +1,9 @@
 // @TODO: These custom property settings seem like they shouldn't need to live
```
callinmullaney, line 1: Per out Tech Debt standards can you create ticket in Jira with these details? Also, add an @link comment here linking it back to the ticket. https://www.notion.so/fourkitchens/Technical-Debt-820b9fda626b4b98bac36f910561767f

callinmullaney (approved):
@ModulesUnraveled Thanks for exhausting all avenues in regards to the `!important` - While I wish we found a better way I think your experimenting showed a rare use case for `!important`. I don't think this should be considered tech debt and the comment seems verbose enough for future devs.

See my other comment in regards to the tech debt you identified and make sure that gets updated before merging. Approved!

kara-franco:
Quick questions I had after our Zoom convo yesterday re: use of !important on active states

- What will be the consequence when a dev would like to create a custom style and change the active state styling? 
- For dark mode/theme the button active color may be different, how will we handle that?

ahitson (approved):

miketullo95 (approved):
looks great! The new animation for wipe looks good!

components/01-atoms/links/_link.scss
ModulesUnraveled: Done

ModulesUnraveled:
@kara-franco I don't think your first question is a concern. This is the design system, and any/every version of button styles should be accounted for here. If there needs to be a new style, it should be added to the system. So there isn't really a situation where someone would "create a custom style and change the active state styling".

As far as dark mode etc. support, I updated the styles to use custom properties instead of defined colors etc. so that those can be modified to support the various color schemes or an eventual dark mode

PR #28, opened 2022-02-19
YALB-76: List

ModulesUnraveled:
### Description of work
- Initial List component. (to support menus)

### Testing Link(s)
- [ ] Navigate to the [Unordered List Story](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/atoms-lists--unordered-list) preview
- [ ] Navigate to the [Ordered List Story](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/atoms-lists--ordered-list) preview

### Functional Review Steps
- [ ] Verify the lists show up

### Design Review
- [x] Verify the designs match the [Figma Designs](https://www.figma.com/file/OVfmCW5s1gI7xnrM8ibX2y/UI-Kit-Alpha-%5BYaleSites%5D)

### Accessibility Review
- [x] Verify the component meets Accessibility requirements

## Notes:
- This doesn't currently support nested items. That'll come later if/when needed. The majority of lists are in a wysiwyg, and therefor don't even use this component. Menus do use this, but have their own mechanism for doing nested list items. So this is just to get the basic pieces in place.

### DIFF
--- a/components/01-atoms/lists/_list-item.twig
+++ b/components/01-atoms/lists/_list-item.twig
@@ -1,0 +1,15 @@
+{#
+ # Available variables:
+ # - list__item__content - the content of the list_item (typically text)
+ #
+ # Available blocks:
+ # - list__item__content - used to replace the content of the list_item with something other than text
+ #   for example: to insert the image and/or link components
+ #}
+{% set list__item__base_class = list__item__base_class|default('item') %}
+
+<li {{ bem(list__item__base_class, list__item__modifiers, list__item__blockname) }}>
+  {% block list__item__content %}
+    {{ list__item__content }}
+  {% endblock %}
+</li>
--- a/components/01-atoms/lists/_list-item.twig
+++ b/components/01-atoms/lists/_list-item.twig
@@ -7,8 +7,11 @@
  #   for example: to insert the image and/or link components
  #}
 {% set list__item__base_class = list__item__base_class|default('item') %}
+{% set list__item__attributes = {
+  class: bem(list__item__base_class, list__item__modifiers, list__item__blockname),
+} %}
 
-<li {{ bem(list__item__base_class, list__item__modifiers, list__item__blockname) }}>
+<li {{ add_attributes(list__item__attributes) }}>
   {% block list__item__content %}
     {{ list__item__content }}
   {% endblock %}
--- a/components/01-atoms/lists/_list.scss
+++ b/components/01-atoms/lists/_list.scss
@@ -1,0 +1,5 @@
+@mixin reset {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
--- a/components/01-atoms/lists/list.twig
+++ b/components/01-atoms/lists/list.twig
@@ -1,0 +1,20 @@
+{#
+ # Available variables:
+ # - list__items - the array of list items
+ #
+ # Available blocks:
+ # - list__content
+ #}
+{% set list__base_class = list__base_class|default('list') %}
+{% set list__type = list__type|default('ul') %}
+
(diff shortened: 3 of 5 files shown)


components/01-atoms/lists/_list.scss
```diff
@@ -0,0 +1,5 @@
+@mixin reset {
```
callinmullaney, line 1: I suggest renaming this to `list-reset` or something a bit more descriptive.

components/01-atoms/lists/_list-item.twig
```diff
@@ -0,0 +1,15 @@
...
+ # Available variables:
+ # - list__item__content - the content of the list_item (typically text)
+ #
+ # Available blocks:
+ # - list__item__content - used to replace the content of the list_item with something other than text
+ #   for example: to insert the image and/or link components
+ #}
+{% set list__item__base_class = list__item__base_class|default('item') %}
+
+<li {{ bem(list__item__base_class, list__item__modifiers, list__item__blockname) }}>
```
callinmullaney, line 11: While I can't think of a good example of where/why one would do this I think we should refactor this to use the `add_attributes(li__attributes)` method instead. This allows the `<li>` to be easily extended in the future (if needed) as well as consistency with the already established components.

components/01-atoms/lists/list.twig
```diff
@@ -0,0 +1,20 @@
...
+ # Available variables:
+ # - list__items - the array of list items
+ #
+ # Available blocks:
+ # - list__content
+ #}
+{% set list__base_class = list__base_class|default('list') %}
+{% set list__type = list__type|default('ul') %}
+
+<{{ list__type }} {{ bem(list__base_class, list__modifiers, list__blockname) }}>
```
callinmullaney, line 11: See comment above regarding `add_attributes(list__attributes)`

callinmullaney (changes requested):

callinmullaney (commented):
@ModulesUnraveled I agree with you that the majority of list instances will not use this component directly so there isn't much of a need to account for nesting features at this level. That being said, do you think it's worth adding some sort of styling to accommodate for some of the nested design choices? Specifically the bullet styles `list-style-type: lower-alpha;` ..etc.

components/01-atoms/lists/_list.scss
callinmullaney: discussed in zoom. The usage of this mixin is namespaced so the result would be `@include list.reset;` which is descriptive.

components/01-atoms/lists/_list-item.twig
ModulesUnraveled: I'm still not convinced this is necessary. The `bem` function exists, and is a simpler implementation that `add_attributes`. So, in my opinion, if you're only setting the `class` attribute, you should use the simpler method. Refactoring isn't hard, and switching to `add_attributes` later, if a need arrises would be my preferred plan of action. But it's not a big deal here, so I went ahead and switched them both.

ModulesUnraveled:
@callinmullaney I have another ticket for lists, so I'll handle the styles there, and this one was really just so that I could break something out of the giant menu PR that I'm working on (in hopes to make review easier and more focused). So all I really need from this is the markup to work so that the menu can implement it.

callinmullaney (approved):
@ModulesUnraveled I hear you but I respectfully disagree and would lean on the extensible option for two reasons.
1. Refactoring at this level later on in the project would require more testing because other components may be implementing this list component and would also need to be checked. This might not be a bunch of extra time but there would be more things to consider when making this change later on.
2. This is a convention we're using for other components and it seems odd to have a one-off that doesn't follow it.

At any rate, I really appreciate you taking the feedback to heart and making these changes now. This is ready to 🚢 from my perspective.

PR #30, opened 2022-02-24
Mobile Header Base

ModulesUnraveled:
### Description of work
- [YALB-86: Utility Nav](https://yaleits.atlassian.net/browse/YALB-86) is temporarily blocked by some design discussions, so this PR abstracts some of the "shared" mobile header and menu stuff so that I can work on breadcrumbs and the main menu without being blocked by the utility nav. So, this includes:
- Menu base templates (not demoed anywhere, since they'll be used by each menu, but don't do anything by themselves)
- Mobile Menu toggle
- Updates to the site-header to support mobile designs

### Testing Link(s)
- [ ] Navigate to the [Full Page Layout](https://deploy-preview-30--dev-component-library-twig.netlify.app/?path=/story/page-layouts-page-layouts--full-width) preview

### Functional Review Steps
- [ ] Verify everything works as expected on small screens

### Design Review

### Accessibility Review

### DIFF
--- a/components/01-atoms/buttons/_buttons.scss
+++ b/components/01-atoms/buttons/_buttons.scss
@@ -10,6 +10,15 @@ $button-outline-weights: (
   '4': var(--border-thickness-4),
 );
 
+// The button-reset mixin is used to remove browser default styles for button
+// elements that shouldn't look like a traditional "button".
+@mixin button-reset {
+  border: 0;
+  background: transparent;
+  padding: 0;
+  color: inherit;
+}
+
 @mixin button {
   --color-button-outline: var(--color-link);
 
@@ -42,7 +51,7 @@ $button-outline-weights: (
 
   // Outline weights options
   @each $weight, $value in $button-outline-weights {
-    &[data-outline-weight='#{$weight}'] {
+    &[data-button-outline-weight='#{$weight}'] {
       --border-thickness-button: #{$value};
     }
   }
--- a/components/01-atoms/buttons/_buttons.scss
+++ b/components/01-atoms/buttons/_buttons.scss
@@ -1,3 +1,5 @@
+@use '../../00-tokens/index' as base;
+
 $button-radii: (
   'soft': var(--button-radius-soft),
   'pill': 100vmax,
@@ -19,10 +21,7 @@ $button-outline-weights: (
   font-weight: var(--font-weights-mallory-500);
   padding: var(--size-spacing-4) var(--size-spacing-7);
   text-decoration: none;
-
-  @media (prefers-reduced-motion: no-preference) {
-    transition: 200ms ease-out;
-  }
+  text-align: center;
 
   &:hover {
     background-color: var(--color-button-bg-hover);
@@ -77,7 +76,13 @@ $button-outline-weights: (
   }
 
   // Hover styles
+  &[data-button-hover-style='fade'] {
+    @include base.animate('color, background-color');
+  }
+
   &[data-button-hover-style='rise'] {
+    @include base.animate('transform, box-shadow');
+
     &:hover {
       transform: translateY(-0.25em);
       color: var(--color-button-text);
@@ -87,7 +92,7 @@ $button-outline-weights: (
   }
 
   &[data-button-hover-style='wipe'] {
-    transition-duration: 400ms;
+    @include base.animate($duration: 400ms);
 
     &:hover {
       &[data-button-style='outline'] {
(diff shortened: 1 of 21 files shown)


kara-franco (changes requested):
Mobile menu toggle: The ARIA states are not being announced to screen readers (expanded, collapsed, pressed etc). 

Keyboard controls are great though! I can open and close the toggle with space, enter; and close with esc.

miketullo95:
@ahitson what is your opinion on having the primary navigation centered on mobile even when you have right or left aligned selected. It might be hard for a user to tap right right/left-most item if its that close to the screen edge. 

image

ModulesUnraveled:
@miketullo95 the primary nav will look very different from this on mobile. What's there now is just a placeholder block. I don't think you'll have a concern once it's actually built  (hopefully next week)

ModulesUnraveled:
@kara-franco I think I've set the button aria attributes appropriately.

ahitson (approved):

components/01-atoms/buttons/_buttons.scss
```diff
@@ -8,6 +10,15 @@ $button-outline-weights: (
   '4': var(--border-thickness-4),
 );
 
+// The button-reset mixin is used to remove browser default styles for button
+// elements that shouldn't look like a traditional "button".
+@mixin button-reset {
```
callinmullaney, line 15: Similar to `list-reset` comment I had earlier. Wouldn't this be namespaced during implementation to `button.reset`? On my local I was able to:
1. Change this to `@include reset  { ....` 
2. Add `@use '../../../01-atoms/buttons/buttons' as button;` at the top of the `_menu-toggle.scss` file
3. Swap out the mixin reference to `@include button.reset;`

components/01-atoms/buttons/button.twig
```diff
@@ -40,8 +41,15 @@
   }) %}
 {% endif %}
 
-<button {{ add_attributes(button__attributes) }}>
+{% if button__href %}
```
callinmullaney, line 44: This looks great! Easy to use a `<a>` element here when needed.

components/01-atoms/buttons/_buttons.scss
ModulesUnraveled: I did this in the Breadcrumbs PR when I refactored all of the `@import`s to `@use`s. https://github.com/yalesites-org/component-library-twig/pull/31/files#diff-e6b4386e20a680dd5b7e303065b8387277192ebb6940cc14cfe6d3a3b1e1f113

callinmullaney (approved):
@ModulesUnraveled looks good. Approved.

kara-franco:
@ModulesUnraveled 
The expanded state continues to not be announced by a screen reader. My hunch is that the aria-pressed state is overriding the expanded announcement (and upon further reading, aria-pressed should only be used for true toggle buttons). 

**I suggest removing the aria-pressed attribute, as well as the aria-controls** (I did more reading and this ARIA attribute has poor support and may not be interpreted the same across assistive tech). 

Further reading: 

- [https://a11ysupport.io/tests/tech__aria__aria-controls](https://a11ysupport.io/tests/tech__aria__aria-controls)
- [https://heydonworks.com/article/aria-controls-is-poop/](https://heydonworks.com/article/aria-controls-is-poop/) (as of JAWS 2018, exposure of aria-controls has been removed). 
- [https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/pressed/#do-not-use-for-marking-expandability](https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/pressed/#do-not-use-for-marking-expandability)
- [https://adrianroselli.com/2019/08/under-engineered-toggles-too.html](https://adrianroselli.com/2019/08/under-engineered-toggles-too.html) (nice write up of how to use aria-pressed for toggle buttons)

ModulesUnraveled:
@kara-franco I've removed those aria attributes. Let me know if it's working as expected now.

Also, can you elaborate (here or in Teams) on what you mean by "aria-pressed should only be used for true toggle buttons"

kara-franco:
@ModulesUnraveled 
Awesome, the menu button is announcing expanded/collapsed now in JAWS and VO. 

As for a true toggle button: I mean a button that switches between two states, off and on vs a menu button. The menu button is following a disclosure pattern, where content is hidden and then exposed from the interaction. At first glance, they seem the same, but the outcome of the interaction is announced differently to blind people. 

- toggle button with dark mode or light mode using aria-pressed: announced as selected 
- hamburger menu button that opens a list of nav links using aria-expanded: announced as expanded, announced as collapsed once closed.

kara-franco (approved):

PR #31, opened 2022-02-27
YALB-75: Breadcrumbs Component

ModulesUnraveled:
### Description of work
- Updates sass `import`s to `use`s
- Creates the breadcrumbs component

### Testing Link(s)
- [ ] Navigate to the [Breadcrumbs Story](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/organisms-menus--breadcrumbs)
- [ ] Navigate to the [Basic Page Example Story](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/page-examples-page-examples--basic-page)

### Functional Review Steps
- [ ] Verify the breadcrumbs function as expected at all screen sizes

### Design Review

### Accessibility Review


ahitson:
On mobile, the "< page" should be Academic Programs and switch to the link style (as opposed to an inactive link), with the focus on the entire item as one unit.

ModulesUnraveled:
@ahitson thanks for that clarification! I think I've updated it correctly now.

components/02-molecules/menu/_menu-item.twig
```diff
@@ -14,29 +19,44 @@
   {% set item__modifiers = item__modifiers|merge(['with-sub']) %}
 {% endif %}
 
-{% import "@molecules/menu/menu.twig" as menus %}
+{# @TODO: When we build this in Drupal. Verify this works #}
+{# @LINK: https://yaleits.atlassian.net/browse/YALB-352 #}
+{% if item.is_active %}
```
callinmullaney, line 24: I'm fairly certain the variable we're looking for here is `in_active_trail`

components/02-molecules/menu/_menu-item.twig
```diff
@@ -14,29 +19,44 @@
...
+    {% set link__attributes = link__attributes|merge({
+      'aria-current': 'page',
+    }) %}
+  {% endif %}
+{% endif %}
 
 {# Set link component as default #}
 {% set list__item %}
-  {% block list__item__content_block %}
+  {% if breadcrumb and item.is_active %}
```
callinmullaney, line 38: Checking for `breadcrumb` here and above within this file seems less than ideal. For instance, you'll probably need to do the same thing for the active items for pagers.

I think you could probably leave the if statement but just check for `item.in_active_trail`. Then ditch the `item__attributes` array altogether and pass this to the link component below and wrap line 39 with a span `<span add_attributes(link__attributes)>{{- item.title -}}</span>`

components/02-molecules/menu/menu-toggle/_menu-toggle.scss
```diff
@@ -1,4 +1,5 @@
-@use '../../../00-tokens/index.scss' as *;
+@use '../../../00-tokens/tokens' as *;
+@use '../../../01-atoms/buttons/buttons';
```
callinmullaney, line 2: :+1:

components/02-molecules/menu/menu-toggle/_menu-toggle.scss
```diff
@@ -13,7 +14,7 @@
 }
 
 .menu-toggle {
-  @include button-reset;
+  @include buttons.reset;
```
callinmullaney, line 17: nice

components/02-molecules/menu/menu.twig
```diff
@@ -34,15 +34,19 @@
 
 {% set menu__attributes = menu__attributes|merge({
   'id': menu__base_class,
-  'class': bem(menu__base_class, menu__blockname, menu__modifiers, menu__additional_classes),
+  'class': bem(menu__base_class, menu__modifiers, menu__blockname, menu__additional_classes),
   'aria-label': menu__name,
 }) %}
 
+{% if breadcrumb %}
```
callinmullaney, line 41: See comment above. I think we need to avoid checking for specific components. In this case it seems like having a variable `ordered_menu = TRUE` or something would be more generic and applicable for multiple components.

components/02-molecules/menu/menu.twig
```diff
@@ -58,7 +62,7 @@
 
 <nav {{ add_attributes(menu__attributes) }}>
```
callinmullaney, line 63: @kara-franco Accessibility question for you. Is it acceptable to use `aria-label` in this instance on the `<nav>` or is it best to use `aria-labelledby="nav-heading"` followed by an `<h2 id="nav-heading">My awesome nav</h2>`

components/03-organisms/site-header/site-header.twig
```diff
@@ -27,7 +27,7 @@
   <div {{ bem('inner', [], site_header__base_class) }}>
     <div {{ bem('mobile-header', [], site_header__base_class) }}>
       {# Yale Branding #}
-      <a href="https://www.yale.edu/" {{ bem('site-branding', ['mobile'], site_header__base_class) }}>
+      <a href="/" {{ bem('site-branding', ['mobile'], site_header__base_class) }}>
```
callinmullaney, line 30: I think this should be a link component? That or maybe add a TODO here to update this to a future `branding` component.

callinmullaney (changes requested):
Let me know if you want to zoom about any of this feedback.

components/02-molecules/menu/_menu-item.twig
ModulesUnraveled: From what I could find `in_active_trail` would apply to the current page _and_ all of it's parents, grandparents, etc. So it's not specifically the "active page". `is_active` is specifically the current page.

ahitson:
Thanks for making those changes. On mobile, will the < be replaced with an icon when we have the library? I also want to put some more padding between the icon and text.

components/02-molecules/menu/menu.twig
ModulesUnraveled: My understanding on this is that if the "label" (in your example the `h2`) already exists, use `aria-labelledby` and point to it. If it doesn't already exist, use `aria-label`. If that's true, I don't think we need to add an extra element (especially since we'd need to visually hide it in this case) to label the nav.

Here's what I read:
https://tink.uk/the-difference-between-aria-label-and-aria-labelledby/

kara-franco: Screen reader users often jump around the page using landmarks, such as navs, so a clear label on the nav will help give them more context/distinguish breadcrumb nav from the main menu nav, let's say. 

aria-label would suffice in this instance! It's easier to read in the code and it is actually supported better across assistive technology.

aria-label support: [https://a11ysupport.io/tech/aria/aria-label_attribute](https://a11ysupport.io/tech/aria/aria-label_attribute) 
aria-labelledby support: [https://a11ysupport.io/tech/aria/aria-labelledby_attribute](https://a11ysupport.io/tech/aria/aria-labelledby_attribute)

components/02-molecules/menu/_menu-item.twig
ModulesUnraveled: Good ideas! I've updated the code per our zoom conversation.

components/02-molecules/menu/menu.twig
ModulesUnraveled: Updated.

components/02-molecules/menu/menu.twig
ModulesUnraveled: Thanks for the clarification, Kara!

ModulesUnraveled:
@ahitson It definitely can be replaced with an icon 👍  I added some space if you can take another look.

@callinmullaney I think everything is ready for you to re-review

callinmullaney (approved):
@ModulesUnraveled code updates and functional look good to me. The `menu__contextual` seems like a more appropriate solution in the generic menu component twig. Also, the cleanup you did with the item_attributes is much appreciated. ready to :shipit: from my perspective.

kara-franco (changes requested):
Mobile view: 
- remove aria-pressed attribute that appears after hitting button (it's interfering with the collapse/expanded announcement).
- add aria-hidden=true to the < (screen reader is saying "less than").

Screen Shot 2022-03-02 at 10 29 11 AM

ahitson (approved):
Wonderful! Let's make a note to remember to add that icon once we have library confirmation and I'll pass this.

kara-franco (approved):

PR #34, opened 2022-03-01
1) YALB-366: Text component

ModulesUnraveled:
### Description of work
- Defines nested list styles
- Creates a text component
- Demos some basic inline text styles (using browser defaults - no custom styles)

### Testing Link(s)
- [x] Navigate to the [Unordered](https://deploy-preview-34--dev-component-library-twig.netlify.app/?path=/story/atoms-lists--unordered-list) and [Ordered](https://deploy-preview-34--dev-component-library-twig.netlify.app/?path=/story/atoms-lists--ordered-list) components
- [x] Navigate to the [Text Examples](https://deploy-preview-34--dev-component-library-twig.netlify.app/?path=/story/atoms-typography--text-examples) 

### Functional Review Steps
- [x] Verify the nested list markers are correct (see screenshot below)
- [ ] Verify the text all looks as expected.

### Design Review

### Accessibility Review

Screen Shot 2022-02-28 at 12 27 25 PM

### DIFF
--- a/components/01-atoms/lists/_list.scss
+++ b/components/01-atoms/lists/_list.scss
@@ -3,3 +3,27 @@
   margin: 0;
   padding: 0;
 }
+
+ul {
+  list-style-type: disc;
+
+  ul {
+    list-style-type: circle;
+
+    ul {
+      list-style-type: square;
+    }
+  }
+}
+
+ol {
+  list-style-type: decimal;
+
+  ol {
+    list-style-type: lower-alpha;
+
+    ol {
+      list-style-type: lower-roman;
+    }
+  }
+}
--- a/components/01-atoms/lists/_list.scss
+++ b/components/01-atoms/lists/_list.scss
@@ -27,3 +27,9 @@ ol {
     }
   }
 }
+
+.text-field {
+  li {
+    padding-left: var(--size-spacing-1);
+  }
+}
--- a/components/01-atoms/lists/list.stories.js
+++ b/components/01-atoms/lists/list.stories.js
@@ -7,5 +7,7 @@ import listData from './list.yml';
  */
 export default { title: 'Atoms/Lists' };
 
-export const UnorderedList = () => listTwig(listData);
-export const OrderedList = () => listTwig({ ...listData, list__type: 'ol' });
+export const UnorderedList = () =>
+  listTwig({ list__items: listData.unordered__list__items });
+export const OrderedList = () =>
+  listTwig({ list__items: listData.ordered__list__items, list__type: 'ol' });
--- a/components/01-atoms/lists/list.stories.js
+++ b/components/01-atoms/lists/list.stories.js
@@ -7,7 +7,13 @@ import listData from './list.yml';
  */
 export default { title: 'Atoms/Lists' };
 
-export const UnorderedList = () =>
-  listTwig({ list__items: listData.unordered__list__items });
-export const OrderedList = () =>
-  listTwig({ list__items: listData.ordered__list__items, list__type: 'ol' });
+export const UnorderedList = () => `
+  <div class="text-field">
+    ${listTwig({ list__items: listData.unordered__list__items })}
+  </div>
+`;
+export const OrderedList = () => `
+<div class="text-field">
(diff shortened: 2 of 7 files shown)


ahitson:
Lists :
- Can we add the smallest # of padding between the numbers and the text? It looks good on unordered but feels too close to the numbers on ordered, so I'd rather push them both out a smidge. 
- Is it possible to thicken the outline on the circle by a px? Decrease the size of the square by a tiny, maybe 10%?

Text examples:
- The line length limit doesn't look to be in effect.
- Does the line-height start at 33 px?
- Will italics be called emphasis or italic text?
- What is the use case for inserted text? The underline is confusing to me.

kara-franco (changes requested):
Awesome, I'm glad you removed the unlikely HTML elements (some of them are not accessible to screen readers!). 

`<s>` strikethroughs are really tricky since they are not communicated to screen reader users. In most cases, the user will completely miss the context of the line (miss out on a sale price/original price for example). 

- MDN offers a solid solution, the only caveat is that it increases verbosity for screen reader users (we need to train our content editors when to use strikethrough and when to not). [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/s](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/s)

Superscripts have another accessibility challenge, where if they are linked (e.g. many academic publications/journal entries) the touch target size is small and the link is not descriptive. Example: Cats are cool<sup>1</sup><sup>2</sup> would need some extra padding to provide easier selection between the "1" and "2" for touchscreen and mouse users. 

- I suggest increasing the font size for the superscript (it is really tiny) and adding a little left-right padding to them.

ModulesUnraveled:
@kara-franco I think the size of the "registered" mark is making it appear extra small. Here's a screenshot with some "regular" text for comparison.
Screen Shot 2022-03-04 at 5 27 39 AM

I think it's worth addressing the padding issue though, even though the text is pretty large. Can you find a link for me with an example of how this is used so that I can see what the "real-world" needs are?

@ahitson Here's a screenshot of adding 2px of padding and 2px of margin between linked items in a superscript context
Screen Shot 2022-03-04 at 5 33 56 AM

kara-franco:
@ModulesUnraveled 
Wow, the actual text size is so much better that the registered icon, lol. 

This page does an exceptional job at making accessible linked footnotes. [https://www.w3.org/WAI/business-case/](https://www.w3.org/WAI/business-case/). What is interesting is that they use an aria-label to provide more context (rather than the link be called "1" they use the label "Footer note 1".

components/01-atoms/lists/_list.scss
```diff
@@ -3,3 +3,33 @@
...
+  ol {
+    list-style-type: lower-alpha;
+
+    ol {
+      list-style-type: lower-roman;
+    }
+  }
+}
+
+.text-field {
```
callinmullaney, line 31: These styles won't be applied in Drupal as `.text-field` isn't a class that's added by default to formatted text fields. We can discuss via zoom. I don't think this should block merging.

callinmullaney (approved):
@ModulesUnraveled I left a code comment that we can chat about at some point next week. Otherwise, this looks good from a code perspective. 

Design note: Maybe it's just me/my browser but the `<sup>` styles added by `normalize.css` are incredibly small. I have a hard time even recognizing the registered symbol:
sup-text

callinmullaney:
:thinking: now that I look at the cropped screenshot it might just be my monitor size and PPI.

kara-franco (approved):

components/01-atoms/lists/_list.scss
ModulesUnraveled: The `text-field.twig` molecule adds this class. I intend to use that on anything that uses a wysiwyg.

ModulesUnraveled:
@kara-franco @callinmullaney Just for your info, I updated the example text for the `<sup>` to just be text. The `®` character is already small, so when we put it inside the sup, it made it extra small. I think this will be more representative of what a user should expect to see:
Screen Shot 2022-03-04 at 4 02 05 PM

callinmullaney:
@ModulesUnraveled good call. :shipit:

PR #35, opened 2022-03-01
2) YALB-73: Images

ModulesUnraveled:
### Description of work
- Creates demos of known image aspect ratios
- Creates a text molecule that supports "content-width"
- Creates an image molecule that supports various widths and a caption
- Tweaks the pull-quote to adhere to expected width
- Creates a news-article example page

### Testing Link(s)
- [ ] Navigate to the [Image Atom](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/atoms-images--images) to see the known aspect ratios
- [ ] Navigate to the [Content Image Story](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/molecules-image--content-image) to see what an image will look like with a caption in content
- [ ] Navigate to the [Text Field Story](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/molecules-text--text-field)
- [ ] Navigate to the [Standard Page Example](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/page-examples-page-examples--standard-page)
- [ ] Navigate to the [News Article Example](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/page-examples-page-examples--news-article)

### Functional Review Steps
- [ ] Verify all of the components make sense :)

### Design Review

### Accessibility Review


kara-franco (changes requested):
Looking great! 

Accessible Figcaptions: 
I suggest we add an aria-label to the` <figure> `that contains the content of the figcaption. There is mixed support of how screen readers handle figcaptions and the aria-label will cover those gaps. I have a link below that shows the method/example and provides (a lot) of explanation about the why to do this! 

More Drupal related: We also need to require the alt text on images in figures. 

https://www.scottohara.me/blog/2019/01/21/how-do-you-figure.html#:~:text=Wrapping%20up,audience%20as%20possible%3A

ModulesUnraveled:
@kara-franco Good to know! I've added the aria-label

kara-franco (approved):

kara-franco:
Perfect, thank you for the label update!

miketullo95:
@ModulesUnraveled does the image atom take into account the responsiveness when images like the example page go down to about 400px wide? I'm not sure if this would be at the image atom level, or a container that they sit in. However, at this breakpoint,  we may want to set display as block at this size. 

Screen Shot 2022-03-04 at 2 57 55 PM

components/01-atoms/images/image/figure.twig
```diff
@@ -0,0 +1,32 @@
...
+} %}
+
+{% if figure__caption %}
+  {% set figure__attributes = figure__attributes|merge({
+    'aria-label': figure__caption,
+  }) %}
+{% endif %}
+
+<figure {{ add_attributes(figure__attributes) }}>
+  {% if image__url %}
```
callinmullaney, line 15: @kara-franco here we have the ability to link the `` within the figure component. Should this `<a>` be required to have a title attribute or is the required alt attribute on the `` make the link self-descriptive enough?

callinmullaney (commented):
@ModulesUnraveled this looks pretty good. I left a comment for you and Kara to consider.

components/01-atoms/images/image/_picture.twig
```diff
@@ -0,0 +1,13 @@
+{% set picture__base_class = picture__base_class|default('picture') %}
+
```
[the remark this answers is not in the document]
ModulesUnraveled: I'm torn on this. We definitely could, but `<picture>` is rarely used... and the actual markup will more likely come from Drupal, so we won't be defining it ourselves.

I don't think I want to add that complexity at this point (but I also don't want to remove the `<picture>` functionality, since it may be used.) So, unless you have strong feelings, I'd like to leave this part as-is for now.

ModulesUnraveled:
@miketullo95 Sometimes they may need to be block, but more often than not, the container will be `display: flex` so the image will be sized based on the container. Keep checking the components as they come to make sure they're all working as expected, but I don't think there's anything to change here.

components/01-atoms/images/image/_picture.twig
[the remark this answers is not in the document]
callinmullaney: @ModulesUnraveled it isn't hurting anything by having it in but it's also not very useful if you aren't showing it. As you always say, keep your code as clean as possible. I don't have strong feelings here so it's ultimately your call.

ahitson:
Not related to images, but the space between the footer and content should be more extreme, I was using --size-spacing-12

ahitson (approved):

miketullo95 (approved):

components/01-atoms/images/image/figure.twig
callinmullaney: @ModulesUnraveled I'll approve once we get confirmation from Kara regarding my accessibility question above.

ModulesUnraveled:
@ahitson I made a note about the space between the footer and content. So it won't get lost, but I'll handle it in another PR.

components/01-atoms/images/image/figure.twig
callinmullaney: Thanks for the info! TIL

callinmullaney (approved):
@ModulesUnraveled :shipit:

PR #2, opened 2022-01-23
2) CLI Support

ModulesUnraveled:
### Description of work
- Creates the initial system.emulsify.json to support the CLI
- Defines the `test` script that can be run locally, and is run during the build process
- Protected the `develop` and `main` branches, making sure the test script succeeds before anything can be merged

### Functional testing steps:
- [x] View the [failing test](https://github.com/yalesites-org/component-library-twig/runs/4921285796?check_suite_focus=true) (where I forced an error through) to verify that the test script will fail the CI workflow
- [x] Install the system locally using the CLI, and verify there are no issues/errors along the way
	- [x] Create a temporary theme and cd into it `emulsify init "Test Yale Theme" --platform drupal . && cd test_yale_theme`
	- [x] Install the yale system (component library) `emulsify system install --repository https://github.com/yalesites-org/component-library-twig.git  --checkout emulsify-cli`
	- [x] Install dependencies and run storybook `npm install && npm run storybook`
	- [x] View the storybook and verify there are colors and spacing
	- [x] Install the divider component `emulsify component install divider`
	- [x] Restart storybook and verify the divider component is available inside StoryBook

PR #37, opened 2022-03-03
3) YALB-368: Pop-out Image component

ModulesUnraveled:
### Description of work
- Created Pop-out image component

### Testing Link(s)
- [ ] Navigate to the [Pop-out Image Story](https://deploy-preview-37--dev-component-library-twig.netlify.app/?path=/story/molecules-pop-out-image--pop-out-image)
- [ ] Navigate to the [News Page](https://deploy-preview-37--dev-component-library-twig.netlify.app/?path=/story/page-examples-page-examples--news-article)

### Functional Review Steps
- [ ] Verify the pop-out image component works at all screen sizes.

### Design Review

### Accessibility Review


kara-franco:
Nice! Two quick questions: 

1. The aria-label will be applied to the `<figure>` once the Image PR is merged, correct? 
2. Will a regular `` be allowed to be used in this style? A user may want to add a decorative image that does not need a <`figcaption`>.

ModulesUnraveled:
@kara-franco
1. Yes, once the image PR is merged the `aria-label` will apply to this component as well.
2. Right now the `figcaption` is not required. If it not provided the image will still be wrapped in the `<figure>` tag, but the only contents will be the image (with its alt text). If we need to remove the figure wrapper when there's no caption, I can do that too. I was just assuming it'd be okay to keep it. Let me know if that was a wrong assumption.

miketullo95:
Is there a caption character limit? and if there is, is that applicable for storybook controls?
Screen Shot 2022-03-04 at 3 17 06 PM

kara-franco:
> @kara-franco
> 
> 1. Yes, once the image PR is merged the `aria-label` will apply to this component as well.
> 2. Right now the `figcaption` is not required. If it not provided the image will still be wrapped in the `<figure>` tag, but the only contents will be the image (with its alt text). If we need to remove the figure wrapper when there's no caption, I can do that too. I was just assuming it'd be okay to keep it. Let me know if that was a wrong assumption.

2. We will still need to provide an accessible name to the `<figure>`. [Test 6 from Scot O'hara's testing](https://scottaohara.github.io/testing/figure/) shows that when an image with alt text sits in a figure, the accessible name is not announced. 

So we'll need to bring in the alt text as an aria-label on the figure OR just remove the figure (I think that's the cleaner option for what's rendered, but not sure how hard that would be on your side?).

ModulesUnraveled:
@miketullo95 There isn't any character limit built in to the design component. The Drupal field, however, can have a character limit if necessary. That type of config is usually handled by the back end, so there isn't anything to do for the StoryBook interface.

components/02-molecules/pop-out-image/pop-out-image.stories.js
```diff
@@ -0,0 +1,34 @@
...
+      type: 'string',
+      defaultValue: 'This is the caption for the 16:9 image above.',
+    },
+  },
+};
+
+export const PopOutImage = ({ caption }) => `
+  ${textFieldTwig({
+    text_field__content:
+      '<p>The team hypothesizes that cells form distinct types of filaments, activated and inhibited, with unique structures that fine-tune metabolic activity. What’s new about their approach is they will study protein structures by infrared microscopy rather than fluorescence microscopy. They expect these images to reveal a new localized layer of cellular metabolism that could be used to re-engineer metabolic processes. In the future, these complementary time-resolved imaging and cross-linking approaches could be extended to biological studies of macromolecular assemblies and phase-separated membraneless organelles that have been historically difficult to study.</p>',
```
callinmullaney, line 26: seems like this markup should be captured in a yml file instead of here. Clean this file up a bit.

callinmullaney (changes requested):
@ModulesUnraveled I don't actually see this component in the linked figma design. Is this a variation of the Text with Image? Couple of suggestions here:
1. Would be nice to have some controls here to define a left or right alignment instead of forcing left.
2. If this is a variation of the text with image component I could see this being more efficient as a data attribute than it's own separate component. `data-image-display` with the options of `popout`/`float` or `column` to be more of the side-by-side layout displayed in figma.

components/02-molecules/pop-out-image/pop-out-image.stories.js
ModulesUnraveled: Oh, right. This was added to the UI kit, so it's not in the original files. I'll attache a screenshot below for reference.

1) I moved the demo content into a yml file.
2) This is not a variation of the text with image component. Mainly because they have very different content requirements. The text with image will have a separate heading field, and character limits on the text, and a separate link field. This component is an image next to a generic "body" wysiwyg field. (no separate heading/link fields, and unlimited text)
3) There isn't anything in the designs to suggest it should be allowed to float right. @ahitson do you have any opinions on whether this should be allowed to float left or right? Or just left?

Screen Shot 2022-03-05 at 5 57 09 AM

callinmullaney: @ModulesUnraveled @ahitson Sounds good that this is a representation of an image within a WYSIWYG. We can scratch the other suggestions I had above. That being said, unless the plan is to remove `align-right` format option from all WYSIWYG profiles we should account for that here as well IMO. 

Otherwise, yml changes look good and I'll approve.

callinmullaney (approved):

components/02-molecules/pop-out-image/pop-out-image.stories.js
ModulesUnraveled: One thing to note is that this isn't actually an image inside a wysiwyg. This is in a "Pop-out Image" component that has an image field, and a wysiwyg field (just for text). So there won't be any alignment options needed.

kara-franco (changes requested):
All good, as long as we have a solution to #2 in this comment: 
[https://github.com/yalesites-org/component-library-twig/pull/37#issuecomment-1059558329](https://github.com/yalesites-org/component-library-twig/pull/37#issuecomment-1059558329)

ahitson (approved):

ModulesUnraveled:
@ahitson as discussed in Slack, I've created this ticket to solve the spacing issue once Jim is back. https://yaleits.atlassian.net/browse/YALB-393

miketullo95 (approved):

PR #39, opened 2022-03-04
5) YALB-80: Callout component

ModulesUnraveled:
### Description of work
- Creates the Callout component
- Adds everything from Figma to the example pages
- Refactors the divider to follow established methodologies

### Testing Link(s)
- [ ] Navigate to the [Callout Story](https://deploy-preview-39--dev-component-library-twig.netlify.app/?path=/story/molecules-callout--callout)
- [ ] Navigate to the [Standard Page](https://deploy-preview-39--dev-component-library-twig.netlify.app/?path=/story/page-examples-page-examples--standard-page)

### Functional Review Steps
- [ ] Verify the callout looks appropriate at all screen sizes

### Design Review

### Accessibility Review


callinmullaney (changes requested):
@ModulesUnraveled looks like the `data-component-theme` isn't being applied properly. Maybe it's a token issue? The callouts on my local and the one deploy to https://deploy-preview-39--dev-component-library-twig.netlify.app/?path=/story/molecules-callout--callout is lacking background colors.

ModulesUnraveled:
@callinmullaney Woops! I forgot to ship the updated tokens for this component. Should be good now.

ModulesUnraveled:
@ahitson Currently, this is just one headline, text, and a link. (There's no subhead.) I did this initially based on OPAC's feedback about stacked headlines. Let's figure out whether this should be updated/handled differently

ahitson:
- Will the button type follow the primary button choice and not always be solid? Should that be a part of the controls?
- The line length on both the heading and body text is too long on the single call out. I'd half it at max view size.
- I prefer that we have strict character limits to the body text so that it remains centered and acceptable. 
- Really not into the headline alignment path and would like to work toward another solution.
- Subheadings were intended to be an option and are in the UI kit. If it's decided that they aren't needed, we can remove them.

kara-franco:
Yay! These sound great with a screen reader. I noticed when zooming in to 400% at 1280px viewport AND on mobile, the line length of the body text / heading text is quite narrow. For easier reading for low-vision users, I suggest using up some of that real estate and pushing out the margins a bit. @ahitson may have some helpful design related advice/insight! 

Screen Shot 2022-03-08 at 9 33 04 AM

miketullo95:
it looks like on the gray background for callouts, the CTA button text has a blue color. 

Screen Shot 2022-03-08 at 11 12 25 AM

ModulesUnraveled:
Okay, first, here's the feedback from April in Jira:
> 1. We would like to add character limits for both the heading and description to two lines of text on large desktop view (we understand they might go to more lines of text on smaller screens). 
> 2. We noticed for the two callouts once we edited text one of the callouts width go smaller - these should always be 50% - 50% screen split until stacked on smaller screen.
> 3. I reviewed with Kara and with the two line character limits for header and body will allow us to center only the text all the time.

1. For the character limit, that will be set on the backend (in Drupal) but what is in Storybook by default should reflect that.
2. I fixed the "not-50%" issue
3. We've centered the text 👍 

Now, @ahitson I think we addressed all of your requests on Zoom, please double check.

@kara-franco I removed a double-padding thing, so there's more space on small screens (and/or when zoomed in)

@miketullo95 Yes, the button text color is "Yale Blue". if you take a look at the [Color Pairings page](https://deploy-preview-39--dev-component-library-twig.netlify.app/?path=/story/tokens-colors--color-pairings), you'll see that the buttons use the Yale blue color for all of the color schemes (including gray-800 and gray-700 - which is what this component uses)

kara-franco (approved):
Very nice, the gutter/side margins in mobile and/or when zoomed in look fantastic, much better.

ahitson (approved):

miketullo95 (approved):

components/05-page-examples/standard-page.twig
```diff
@@ -9,4 +9,24 @@
...
     } %}
+    {% include "@atoms/divider/divider.twig" %}
+    {% include "@molecules/text/text-field.twig" with {
+      text_field__content: '<h2>Program Inforamtion</h2><h3>The Undergraduate Handbook</h3><p>A comprehensive guide for prospective and current Chemistry Majors, with a complete description of requirements and opportunities.</p><p><a href="#">View the handbook online</a></p><h3>Course List</h3><p>A list of courses and a description of the Chemistry undergraduate program, including information on placement exams, laboratory registration, premedical students, and major requirements</p><p><a href="#">Yale College programs of study</a></p>',
```
callinmullaney, line 14: Not a blocker for this PR but we should probably move all this extra twig/storybook data to it's own file. As the project progresses these static values are really going to muddy up this file.

callinmullaney (commented):
Code changes look good to me. I left a non-blocking comment on the standard-page.twig file. Otherwise, I do want to point something out that I don't believe was mentioned.

When viewing the standard page the variable string lengths case line-breaks at different sizes. While doing a character limit (mentioned about) will help this, it wont solve the problem. What I'm seeing is the CTA/buttons being displayed at various heights. Design/layout-wise this feels off to me.

components/05-page-examples/standard-page.twig
ModulesUnraveled: We talked about the position of the button, and decided to keep it up "next to" the content instead of pushing it down. There are various screen sizes and content lengths that affect the amount of space that is created, and it was decided that keeping it close was preferred.

And about keeping content out of the page-examples... Evan always said the same thing, but I prefer to keep it in the twig files for example pages. We can talk it through if you ever want to, but I've found that when there are multiples of the same component on a page (like two text fields) you have to namespace them (`text__content: text__content__1` and `text__content: text__content__2` etc.) in order to get the right content in the right components. Just putting it in the file means you don't have to do that extra mapping... and you don't get a cluttered folder (two files per page instead of one). I think this is a personal preference thing, and I'm going to keep it like it is for now, but if it does ever seem like it's getting overwhelming, it'll be an easy refactor.

callinmullaney: Sounds good regarding the design. Figured I would mention it.

hmmm Evan was correct. RIP.

callinmullaney (approved):
@ModulesUnraveled thanks for the response. :shipit:

PR #42, opened 2022-03-09
YALB-108: Standard Page

ModulesUnraveled:
### Description of work
- Adds into content examples to standard page
- Sets the page-title width to "content" width
- Updates tokens to remove "beige" from site-footer color schemes

### Testing Link(s)
- [ ] Navigate to the [Standard Page Story](https://deploy-preview-42--dev-component-library-twig.netlify.app/?path=/story/page-examples-page-examples--standard-page)

### Functional Review Steps
- [ ] Verify the Standard Page story has a new "Intro Content" control, and you can choose "image", "image-feature", "image-max", and "pop-out-image" to see each of them at the top of the page.

### Design Review

### Accessibility Review

### DIFF
--- a/components/02-molecules/menu/menu.twig
+++ b/components/02-molecules/menu/menu.twig
@@ -33,7 +33,6 @@
 {% endif %}
 
 {% set menu__attributes = menu__attributes|merge({
-  'id': menu__base_class,
   'class': bem(menu__base_class, menu__modifiers, menu__blockname, menu__additional_classes),
   'aria-label': menu__name,
 }) %}
--- a/components/02-molecules/pop-out-image/_pop-out-image.scss
+++ b/components/02-molecules/pop-out-image/_pop-out-image.scss
@@ -16,3 +16,7 @@
     margin-bottom: var(--size-spacing-7);
   }
 }
+
+.pop-out-image__clearfix {
+  clear: both;
+}
--- a/components/02-molecules/pop-out-image/pop-out-image.twig
+++ b/components/02-molecules/pop-out-image/pop-out-image.twig
@@ -41,3 +41,4 @@
     } %}
   </div>
 </div>
+<div {{ bem('clearfix', [], pop_out_image__base_class) }}></div>
--- a/components/03-organisms/site-header/_site-header.scss
+++ b/components/03-organisms/site-header/_site-header.scss
@@ -5,6 +5,7 @@
 $site-header-themes: map.deep-get(tokens.$tokens, site-header-themes);
 $site-header-border-thickness: map.deep-get(tokens.$tokens, border, thickness);
 $site-header-layout: map.deep-get(tokens.$tokens, 'site-header-layout');
+$site-header-animation-speed: 800ms;
 
 // Mixin for the small branding typography.
 @mixin branding-small {
@@ -65,7 +66,7 @@ $site-header-layout: map.deep-get(tokens.$tokens, 'site-header-layout');
   flex-direction: column;
 
   @media (max-width: tokens.$break-mobile-max) {
-    @include tokens.animate(max-height, 800ms);
+    @include tokens.animate(max-height, $site-header-animation-speed);
 
     z-index: 1;
     position: absolute;
@@ -78,6 +79,22 @@ $site-header-layout: map.deep-get(tokens.$tokens, 'site-header-layout');
 
     [data-main-menu-state='closed'] & {
       max-height: 0;
+
+      // This animation is to remove hidden items from screen readers and
+      // keyboard navigation, while maintaining the slow "close" effect.
+      visibility: hidden;
+      animation: $site-header-animation-speed fade-out;
+      animation-fill-mode: forwards;
+
+      @keyframes fade-out {
+        99% {
+          visibility: visible;
+        }
+
+        100% {
+          visibility: hidden;
+        }
+      }
     }
   }
 }
@@ -192,12 +209,13 @@ $site-header-layout: map.deep-get(tokens.$tokens, 'site-header-layout');
 
   // The animate mixin doesn't support multiple transitions.
   @media (prefers-reduced-motion: no-preference) {
-    transition: opacity 800ms, height 0ms ease-out 800ms;
+    transition: opacity $site-header-animation-speed,
+      height 0ms ease-out $site-header-animation-speed;
   }
 
   @media (max-width: tokens.$break-mobile-max) {
     [data-main-menu-state='open'] ~ & {
-      @include tokens.animate(opacity, 800ms);
+      @include tokens.animate(opacity, $site-header-animation-speed);
 
       height: 100vh;
       opacity: 1;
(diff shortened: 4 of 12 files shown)


kara-franco (changes requested):
I'm still seeing the aria-pressed=true after opening the breadcrumb menu in mobile view (this is interfering with the screen reader announcements). Remove the aria-pressed-true after button activation (note: aria-pressed=false is not present before activating the button). 

Screen Shot 2022-03-09 at 11 57 22 AM

Besides that looking awesome!

ModulesUnraveled:
@kara-franco Whoops! That was missed when I refactored the buttons. Should be fixed now.

kara-franco (changes requested):
Awesome, the aria-pressed=true is removed from the breadcrumb now. 

This is the first time I'm testing the navs/mobile navs after that last merge and I caught one other item just now (I must have missed this, womp)! 

The child links (ex. util) under the mobile nav button are exposed in the DOM when the button is closed/collapsed. The links are visually hidden, but also need to be hidden from screen reader users (with aria-hidden). We can then remove the aria-hidden when the mobile menu button is activated/expanded. 

Steps to replicate: 

- Bring browser window size down to display mobile menu hamburger button. 
- Tab with keyboard past the mobile menu button. 
- Note that there are extra hidden tab stops after the menu (even though is it closed). 

Let me know if you need more info! I'm sorry I didn't see this after the last merge, I'm thinking it can be a quick fix with aria-hidden magic🤞🏻. 

Reference pics: 
Screen Shot 2022-03-11 at 9 34 09 AM

Screen Shot 2022-03-11 at 9 44 55 AM

ModulesUnraveled:
@kara-franco based on our Teams chat, this is going to be a little tricky to make sure things are properly hidden for screen reader and keyboard navigating user, but not "flicker" for sighted users. With that in mind, would it be okay with you if I address this in the primary nav ticket (that I started this morning) so that the two PRs in here currently can get merged?

kara-franco:
> @kara-franco based on our Teams chat, this is going to be a little tricky to make sure things are properly hidden for screen reader and keyboard navigating user, but not "flicker" for sighted users. With that in mind, would it be okay with you if I address this in the primary nav ticket (that I started this morning) so that the two PRs in here currently can get merged?

Sounds like a plan!

kara-franco (approved):
We will address the keyboard/SR issue with the nav in the Primary Nav ticket.

miketullo95 (approved):

ahitson (approved):

components/02-molecules/pop-out-image/_pop-out-image.scss
```diff
@@ -16,3 +16,7 @@
     margin-bottom: var(--size-spacing-7);
   }
 }
+
+.pop-out-image__clearfix {
```
callinmullaney, line 20: I see why we need a clearfix here. I'd be interested in hearing your thoughts but I typically set clearfixes with an `:after` to avoid any `:last-child` styles and to have a cleaner DOM.

components/02-molecules/pop-out-image/pop-out-image.twig
```diff
@@ -41,3 +41,4 @@
     } %}
   </div>
 </div>
+<div {{ bem('clearfix', [], pop_out_image__base_class) }}></div>
```
callinmullaney, line 44: Should this be inside the inner or at least the main `pop-out-image` wrapper so it's closer to the floated image? I guess this works but it seems odd that it's on the very outside.

components/03-organisms/site-header/_site-header.scss
```diff
@@ -78,6 +79,22 @@ $site-header-layout: map.deep-get(tokens.$tokens, 'site-header-layout');
 
     [data-main-menu-state='closed'] & {
       max-height: 0;
+
+      // This animation is to remove hidden items from screen readers and
+      // keyboard navigation, while maintaining the slow "close" effect.
+      visibility: hidden;
```
callinmullaney, line 85: Interesting... I never noticed any issues here so I'd like to talk about this so I'm understanding this solution correctly. Nothing to change here before merging :)

components/02-molecules/menu/menu.twig
```diff
@@ -33,7 +33,6 @@
 {% endif %}
 
 {% set menu__attributes = menu__attributes|merge({
-  'id': menu__base_class,
```
callinmullaney, line 36: probably a good idea to remove this and all ID's to be added when needed on a per-menu basis.

callinmullaney (approved):
@ModulesUnraveled I'm approving to merge as I don't see anything that's a blocker here and the FE obviously looks good for demo. Maybe my comments around the clearfix warrant a change before merging.

That being said, when I testing out this PR I noticed several JS errors. 
`Uncaught TypeError: can't access property "addEventListener", breadcrumbsButton is null`
`Uncaught TypeError: can't access property "addEventListener", menuToggle is null`
...etc.

This is likely only a storybook issue as it loads everything but it might be a good idea to wrap a simple `if (menuToggle) {` around these eventListeners to avoid this altogether.

components/03-organisms/site-header/_site-header.scss
kara-franco: Yes! It is such a common issue with hamburger menus that I see. So the problem is that when the hamburger nav is closed, sighted users do not see the content (nav, util, nav, search,  etc). We want the same to hold true for keyboard- 
only and screen reader users. Otherwise, they will be tabbing through content that is positioned way off the page, not understanding the context. 

For example, on the SOM hamburger menu, there is a visibility:hidden on the header__menu wrapper. If I remove/uncheck that CSS property and tab passed the menu when it is closed, I will tab through all the menu options (that are floating off screen). 

[https://som.yale.edu/](https://som.yale.edu/)
Screen Shot 2022-03-15 at 5 17 36 PM

components/02-molecules/pop-out-image/_pop-out-image.scss
ModulesUnraveled: It's been so long since I've done a clearfix, I forgot how to do it. I originally tried as an `::after` and it didn't work, so I just gave up... I should have googled faster because I just needed to add `content: ''` to get it to work 🤦

components/02-molecules/pop-out-image/pop-out-image.twig
ModulesUnraveled: Removed when I added the `::after` clearfix

components/02-molecules/pop-out-image/_pop-out-image.scss
ModulesUnraveled: @kara-franco Can you verify that the `::after` pseudo element doesn't add anything unwanted to assistive tech? I saw some examples that used `visibility: hidden` but they all had a character in the "content" attribute, so I understood it for that... but since I'm using an empty content, I didn't think I would need it.

Can you just verify that though before I merge?

kara-franco: There are no issues with having empty pseudo content! [Screen readers have mixed support](https://www.powermapper.com/tests/screen-readers/content/css-generated-content/) when it comes to interpreting pseudo content, so it's best to only use :after / :before for non-visual content (content that is hidden/empty) or decorative content. Even with decorative content, like icons, it's best to stick that in a `<span>` and use aria-hidden, since some screen readers do pick up CSS generated content (it's not stable though).

PR #46, opened 2022-03-18
2) YALB-84: Primary-nav

ModulesUnraveled:
### NOTES:
- DONE ~The toggle arrows will be replaced with icons once that PR is done. For now, it's a simple caret (so the alignment is off)~
- There are some other known design tweaks Ashley and I will make next week (font sizes, etc) I don't think any of them should prevent other PR reviews from happening though.

### Description of work
- Creates the Primary Nav component

### Testing Link(s)
- [ ] Navigate to the [Primary Nav Story](https://deploy-preview-46--dev-component-library-twig.netlify.app/?path=/story/organisms-menu-primary-nav--primary-nav) to see it in isolation
- [ ] Navigate to the [Any of the example page stories](https://deploy-preview-46--dev-component-library-twig.netlify.app/?path=/story/page-examples-page-examples--standard-page) to see it in context of the site header/page

### Functional Review Steps
- [ ] Verify everything looks and functions as expected above and below 992px

### Design Review

### Accessibility Review


kara-franco (changes requested):
Mobile menu keyboard accessibility is perfect! Also, the menu levels are announced really nicely in VoiceOver and NVDA screen readers. 

Only issue I caught is that the drop down menu remains open when tabbing to the next Level 1 (ex. menu stays open when I tab to Academics). The expected behavior is to close the drop down when the user tabs out of it.  

Screen Shot 2022-03-22 at 6 53 44 PM

kara-franco:
re: @ahitson design change to remove focus indicators on mouse click- there is a nifty polyfill available called focus-visible that can solve this issue relatively easily. We used it in our component library: 

- [https://yale-a11y.gitlab.io/ui-component-library/documentation#version-history-and-changelog#focus-visible](https://yale-a11y.gitlab.io/ui-component-library/documentation#version-history-and-changelog#focus-visible)
- [https://github.com/WICG/focus-visible](https://github.com/WICG/focus-visible)

You may already know this, but wanted to share!

ModulesUnraveled:
@kara-franco Thanks for pointing me in that direction, I've used `:focus` and `:focus-within`, but hadn't really explored `:focus-visible` yet. [It looks like it's supported](https://caniuse.com/?search=focus-visible) (as a selector) in every browser that we support except Safari at this point.

So, I'd recommend that instead of implementing a polyfill, we just use the selector according to the spec, and provide the small fallback mentioned in the polyfill's docs so that for browsers that don't support `:focus-visible` (namely Safari) use browser styles, but everything else will use our "designed" styles.

I just checked on Chrome and Safari 15.3 locally, and both browsers seem to work as expected (Chrome uses our styles, and Safari uses browser styles.)

I'm making more changes now, but I just pushed that up. So, if you want to take a look at that, and let me know if it seems to work in all of your environments, I think that's the way we should go, personally.

kara-franco:
> @kara-franco Thanks for pointing me in that direction, I've used `:focus` and `:focus-within`, but hadn't really explored `:focus-visible` yet. [It looks like it's supported](https://caniuse.com/?search=focus-visible) (as a selector) in every browser that we support except Safari at this point.
> 
> So, I'd recommend that instead of implementing a polyfill, we just use the selector according to the spec, and provide the small fallback mentioned in the polyfill's docs so that for browsers that don't support `:focus-visible` (namely Safari) use browser styles, but everything else will use our "designed" styles.
> 
> I just checked on Chrome and Safari 15.3 locally, and both browsers seem to work as expected (Chrome uses our styles, and Safari uses browser styles.)
> 
> I'm making more changes now, but I just pushed that up. So, if you want to take a look at that, and let me know if it seems to work in all of your environments, I think that's the way we should go, personally.

Awesome!!! So I just updated to the latest version of Safari and it now supports focus-visible, so I think the above solution is perfect. 

I did notice something odd in Firefox with the `<button>` focus-visible styling, it's being overridden with a 1px dotted gray indicator. 

Screen Shot 2022-03-23 at 1 47 11 PM

ModulesUnraveled:
@kara-franco It looks like that gray dotted outline is a FireFox-specific reset that ships with normalize... I tried to get rid of it, but then it broke the styles in other browsers... is it something we can live with? If not, I think we should create a bug-fix ticket to address it later. I don't think it's something I should spin my wheels on at this time.

I think your other issues are addressed.

kara-franco:
> @kara-franco It looks like that gray dotted outline is a FireFox-specific reset that ships with normalize... I tried to get rid of it, but then it broke the styles in other browsers... is it something we can live with? If not, I think we should create a bug-fix ticket to address it later. I don't think it's something I should spin my wheels on at this time.
> 
> I think your other issues are addressed.

@ModulesUnraveled Excellent, I've confirmed on Safari, Firefox and Chrome that the menu closes when tabbed out. 

I completely agree, I will log a bug-fix for the Firefox buttons.

kara-franco (approved):

ModulesUnraveled:
@ahitson I converted your comment to a check list, and addressed all of the items except one:
- As we talked about on Zoom, we'll wait for the new icons before we start fiddling with their sizes.

Ready for your re-review

ahitson:
So close! Please make the space between the parent link (2nd level) and first child (3rd level) equal to the space between the child pages.

On mobile, please bring up the first child in the dropdown closer to the parent.

Can we add a smidge of space between the main level menu items or does that put us in dangerous territory?

And bring down the menu so it's hitting the baseline of the logo (example below)
Frame 456 (1)

components/02-molecules/menu/menu-toggle/menu-toggle.js
```diff
@@ -1,42 +1,135 @@
 Drupal.behaviors.menuToggle = {
   attach(context) {
-    // Selectors
+    // Selectors.
```
callinmullaney, line 3: Per our slack discussion you can likely trim down this file. 

- JS should calculate the the header height but assign that value to a CSS custom property.
- SCSS should use that custom property to determine the height of the mobile menu `height: calc(100vw - var(--heade-height);`
- Remove the media query JS. The frozen class should have no affect on desktop sizes which can be applied via a standard CSS media query

components/03-organisms/menu/primary-nav/_primary-nav.scss
```diff
@@ -0,0 +1,201 @@
...
+
+    &[aria-expanded='true'] {
+      z-index: 1;
+      border-color: var(--color-navigation-border);
+      border-bottom-color: var(--color-background);
+    }
+  }
+}
+
+@mixin primary-nav-item-level-1 {
```
callinmullaney, line 41: Not a blocker for merging but these mixins are only used once and are not very big.

callinmullaney (changes requested):
@ModulesUnraveled back to you

components/03-organisms/menu/primary-nav/_primary-nav.scss
ModulesUnraveled: Done

components/02-molecules/menu/menu-toggle/menu-toggle.js
ModulesUnraveled: I did all of this. 👍

ModulesUnraveled:
@ahitson @callinmullaney Back to you two for reviews!

components/03-organisms/site-header/_site-header.scss
```diff
@@ -66,19 +82,19 @@ $site-header-animation-speed: 800ms;
...
   @media (max-width: tokens.$break-mobile-max) {
-    @include tokens.animate(max-height, $site-header-animation-speed);
+    @include tokens.animate(transform, $site-header-animation-speed);
 
-    z-index: 1;
     position: absolute;
     left: 0;
     background: var(--color-background);
-    max-height: 100vh;
+    max-height: calc(100vh - var(--header-height-mobile));
```
callinmullaney, line 90: This is much more noticeable when the border thickness value is set to `8` but I believe this line should be `max-height: calc(100vh - var(--header-height-mobile) + var(--site-header-border-thickness));`

callinmullaney (approved):
@ModulesUnraveled The JS file looks much better and is doing a lot less compared to before. Nice work! 

There is one last minor comment I left for you to address. Otherwise, this seems good to merge from my perspective.

components/03-organisms/site-header/_site-header.scss
ModulesUnraveled: I updated the css to keep the header border in sight. Hoping that's what you were trying to achieve.

ModulesUnraveled:
@ahitson I made the sub menus "slide" in and out on mobile.

@kara-franco Would you mind double checking the mobile nav now? I had to change some stuff to make the sliding work, but I used a methodology we've used elsewhere, so I think it should be accessible

kara-franco:
Very slick!! I tested in Chrome, Safari and Firefox, looks great, keyboard accessibility is very good. Sounds great in VO.

ahitson (approved):

PR #47, opened 2022-03-21
3) YALB-97: Accordion component

ModulesUnraveled:
### Description of work
- Creates the Accordion component

### Testing Link(s)
- [ ] Navigate to the [Accordion Story](https://deploy-preview-47--dev-component-library-twig.netlify.app/?path=/story/molecules-accordion--accordion)
- [ ] Navigate to the [Accordion Page](https://deploy-preview-47--dev-component-library-twig.netlify.app/?path=/story/page-examples-page-examples--accordion-page)

### Functional Review Steps
- [ ] Verify the accordions each toggle only themselves
- [ ] Verify the expand/collapse all buttons toggle all of the accordions in that group (if multiple groups on a page)

### Design Review

### Accessibility Review


ahitson:
- Please put a pipe and some padding between collapse and expand all triggers.
- Can we remove the focus state sticking around on expansion? I'll only want to see the focus state if I'm using it for navigation.
- Add some large padding to the end of the accordion. Thinking it'll always need some significant space between it and whatever follows.

components/02-molecules/accordion/_accordion-item.twig
```diff
@@ -0,0 +1,29 @@
+{% set accordion__item__base_class = 'accordion-item' %}
+
+{% set accordion__item__button__content %}
+  {{ accordion__item__heading }}
+  {% include "@atoms/images/icons/_icon.twig" with {
+    icon__name: 'angle-down',
+    icon__blockname: accordion__item__base_class,
```
kara-franco, line 7: Add icon__decorative here to stop screen reader from saying "graphic".

components/02-molecules/accordion/accordion.twig
```diff
@@ -0,0 +1,37 @@
...
+  <div {{ bem('inner', [], accordion__base_class) }}>
+    {% include "@atoms/typography/headings/heading.twig" with {
+      heading__level: '3',
+      heading__blockname: accordion__base_class,
+      heading: accordion__heading,
+    } %}
+    {% embed "@atoms/lists/list.twig" with {
+      list__base_class: 'controls',
+      list__blockname: accordion__base_class,
+      list__attributes: {'aria-labels': 'Section controls'},
```
kara-franco, line 13: Change to aria-label

components/02-molecules/accordion/accordion.twig
```diff
@@ -0,0 +1,37 @@
...
+          button__content: 'Expand all',
+          button__base_class: 'toggle-all',
+          button__modifiers: ['expand'],
+          button__blockname: accordion__base_class,
+        } %}
+        {% include "@atoms/buttons/button.twig" with {
+          button__content: 'Collapse all',
+          button__base_class: 'toggle-all',
+          button__modifiers: ['collapse'],
+          button__blockname: accordion__base_class,
```
kara-franco, line 26: Both collapse and expand buttons need aria-expanded/collapsed to be announced. Also, can we combine the interactions into one button (like the Figma design)? 

Screen Shot 2022-03-22 at 8 11 15 PM

kara-franco (changes requested):
Excellent work on the accordions (semantics are great and sound awesome for screen reader users). 

My feedback is to update the aria-labels to aria-label, add icon__decorative to the chevron/down arrow and aria-expanded to the Expand/Collapse buttons. 

We should explore having a toggle button to expand/collapse accordions (that uses aria-pressed).

miketullo95:
> We may want to explore having one button to expand/collapse (like in the designs).

I think having one button for the function of collapsing/expanding all is a great choice. My one question for that would be how would the button/link text change is the user manually opens all the accordions. Would it still be the default "expand all" or would the system acknowledge that the accordions are open and change to "collapse all"?

components/02-molecules/accordion/accordion.twig
kara-franco: I was thinking, can we have ARIA attributes added to our linter? I have misspelled aria 1000x (usually aria-labelledby, lol).

components/02-molecules/accordion/accordion.twig
kara-franco: I just got off a call with Mike V and we chatted about the Expand/Collapse toggle pattern here. We are thinking that a true toggle button will be more clear to sighted and blind users. There are two states that can be turned on or off and the two buttons || one button with two labels does not convey this as clearly as a toggle would. 

Reading about accessible toggles: [https://adrianroselli.com/2019/08/under-engineered-toggles-too.html](https://adrianroselli.com/2019/08/under-engineered-toggles-too.html)

kara-franco:
During my testing, I found that the spacing is a little wonky in Safari

Screen Shot 2022-03-23 at 11 17 35 AM

kara-franco:
> @kara-franco I love the toggle idea! I reached out to April and she agrees. I think it's an interesting point to consider "open/close all" for the text of the toggle. "Expand/collapse" seems technical while open/close may be more direct and easier to understand.

@miketullo95 I agree, Open / Close or Show / Hide are waaay more user friendly!

components/02-molecules/accordion/_accordion-item.twig
ModulesUnraveled: Done

components/02-molecules/accordion/accordion.twig
ModulesUnraveled: Done.

I don't know of any way to check for mis-spellings... If you find something though, it would be worth looking into. (My brief google search didn't find anything)

ahitson:
My vote is for the UK gov version with open all and close all! This example works exactly how I'd like, leaving the button as the open all version until every instance is opened, then switching to close once all have been opened.

ModulesUnraveled:
@ashley @kara-franco @miketullo95 I think I've made all of the changes requested with the exception of changing "Expand/Collapse" to "Open/Close"

Ashely and I were talking, and I mentioned that "Open", to me, indicates something like a modal is going to open. In this case, we're actually just "expanding" the section. There isn't a thing to "open". "show/hide" also seemed to have issues because hide could mean that the items will be literally hidden from the page, which isn't the case (the click target is still there).

If you don't agree that "Open/Close" should be reserved for modal-like components, let's jump on a zoom and discuss. Otherwise, we'll stick to "Expand/Collapse" for this component.

kara-franco (approved):

components/02-molecules/accordion/accordion.twig
```diff
@@ -0,0 +1,53 @@
...
+  {% include "@atoms/images/icons/_icon.twig" with {
+    icon__name: 'angle-down',
+    icon__blockname: accordion__base_class,
+  } %}
+{% endset %}
+
+<div {{ bem(accordion__base_class) }} data-component-width="content">
+  <div {{ bem('inner', [], accordion__base_class) }}>
+    {% include "@atoms/typography/headings/heading.twig" with {
+      heading__level: '3',
```
kara-franco, line 22: - [ ]  `accordion__heading`should be a `<h2>`, since the `accordion-item__headings` are `<h3>`s. This will ensure proper nesting. 

Screen Shot 2022-03-25 at 10 17 24 AM

kara-franco: Also, quick question, how will the `accordion__heading` be filled in?

components/02-molecules/accordion/accordion.twig
```diff
@@ -0,0 +1,53 @@
...
+      heading: accordion__heading,
+    } %}
+    {% embed "@atoms/lists/list.twig" with {
+      list__base_class: 'controls',
+      list__blockname: accordion__base_class,
+      list__attributes: {'aria-label': 'Section controls'},
+    } %}
+      {% block list__content %}
+        {% include "@atoms/buttons/button.twig" with {
+          button__content: expand_all__content,
```
kara-franco, line 33: - [ ] Can we wrap the buttons in `<li>`s? The accessibility checker in Storybook flagged this since buttons cannot be children of `<ul>`s. Even though it sounds great in VO and NVDA screen readers, we should add these in to pass tests/pass HTML validators. We can always make a bug-fix if this requires a re-work, not a blocker. 

Screen Shot 2022-03-25 at 10 34 40 AM

components/02-molecules/accordion/_accordion.scss
```diff
@@ -0,0 +1,94 @@
...
+@use '../../01-atoms/atoms';
+
+// The accordion group.
+.accordion {
+  margin-bottom: var(--spacing-page-inner);
+  padding-block: var(--size-spacing-6);
+}
+
+.accordion__controls {
+  @include atoms.list-reset;
```
callinmullaney, line 11: This is a new way to reference these resets. Example, the breadcrumbs scss file uses the list/button separately, then `@include list.reset;` or `@include buttons.reset;` within the file.

I don't know which option is best but want to point out the inconsistency.

callinmullaney (changes requested):
@ModulesUnraveled I left a couple of comments here. In general I think this component is pretty good for the first pass. I wish the html spec for summary/details allowed for headings within `<summary>` to prevent a good portion of the JS you're introducing here but the accessibility benefits are more important. :man_shrugging:

kara-franco (changes requested):
Yay, everything from my initial review is 100%. 

I took a look at the items flagged in Storybook's accessibility checker and made a few comments on those. No blockers, but small tweaks needed to pass accessibility testing/improve UX for screen reader users.

components/02-molecules/accordion/accordion.twig
ModulesUnraveled: I've got the conditional heading levels done. (also conditionally showing the top-level heading markup only if a heading is provided)

@kara-franco The heading will be filled in when someone places an Accordion on the page.
- They'll be asked if they want to give a name to their accordion grouping
- Then they'll be able to add accordion items

components/02-molecules/accordion/accordion.twig
ModulesUnraveled: Done

components/02-molecules/accordion/_accordion.scss
ModulesUnraveled: Yeah, I wanted to get away from `@use`ing twelve different files for the various mixins and variables I needed. That's why the `@forward` is in each of the levels files `tokens.scss`, `atoms.scss`, etc. 

The breadcrumbs should be updated to use this method. I'll go ahead and do that now, but if you find any others, they should be done in separate tickets.

ahitson (approved):

components/02-molecules/accordion/_accordion.scss
callinmullaney: Sounds good. Thanks for clarifying the intent here.

callinmullaney (approved):
All looks good to me! This is a solid accordion component :fist_right::fist_left:

components/02-molecules/accordion/accordion.twig
kara-franco: Very nice, thank you!

components/02-molecules/accordion/accordion.twig
kara-franco: Perfection!

kara-franco (approved):

PR #53, opened 2022-03-28
2) YALB-400: Wysiwyg text and YALB-369: Typography

ModulesUnraveled:
### Description of work
- Updates the site name and "Yale University" branding font sizes
- Updates line-heights of h3-h5s
- Adds spacing around headings inside wysiwyg fields
- Tweaked header/primary nav spacing and alignment

### Testing Link(s)
- [ ] Navigate to the [Text Field Story](https://deploy-preview-53--dev-component-library-twig.netlify.app/?path=/story/molecules-text--text-field)
- [ ] Navigate to the [any Page example](https://deploy-preview-53--dev-component-library-twig.netlify.app/?path=/story/page-examples-page-examples--standard-page)

### Design Review

### Accessibility Review


ahitson (approved):

components/01-atoms/typography/headings/_headings.scss
```diff
@@ -34,4 +34,14 @@ h6 {
...
     @include tokens.h6-yale-new;
   }
+
+  * + {
+    h2,
+    h3,
+    h4,
+    h5,
+    h6 {
+      margin-top: 1em;
```
callinmullaney, line 44: Should this be `var(--size-spacing-7)` or another token ?

components/01-atoms/typography/headings/_headings.scss
```diff
@@ -7,7 +7,7 @@ h4,
 h5,
 h6 {
   color: var(--color-heading);
-  margin-block: 0;
+  margin-block: 0 0.6em;
```
callinmullaney, line 10: Similar comment as below. We don't have an exact spacing match but `--size-spacing-3` or `--size-spacing-4` may be a better option?

components/01-atoms/typography/headings/_headings.scss
callinmullaney: This might be better tackled as a mixin instead of re-declaring the heading selectors again.

```
@mixin header-spacing {
   &:not(:first-child) {
      margin-top: var(--size-spacing-7);
   }
}
```

callinmullaney (changes requested):
@ModulesUnraveled small bit of feedback here. Otherwise, this is looking good to me.

components/01-atoms/typography/headings/_headings.scss
ModulesUnraveled: In this case, no. We're using `em` because we want the value to adjust relative to the font size. So the larger the font gets, the larger the margin.

components/01-atoms/typography/headings/_headings.scss
ModulesUnraveled: The mixin works for me.

(same comment as below about the value though, it's `em`, nor `rem` so there's no token for that.)

callinmullaney (approved):
@ModulesUnraveled Looks good! Thanks for the explanation and implementing the mixin 🎉

PR #56, opened 2022-04-01
1) YALB-392: CTA/Controls

ModulesUnraveled:
### Description of work
- Refactors link and button into controls
	- link -> text-link
	- button -> cta
- Replaces the temporary `<` in breadcrumbs with an icon

### Testing Link(s)
- [ ] Navigate to the [CTA Story](https://deploy-preview-56--dev-component-library-twig.netlify.app/?path=/story/atoms-controls--cta)
- [ ] Navigate to the [Text Link Story](https://deploy-preview-56--dev-component-library-twig.netlify.app/?path=/story/atoms-controls--text-link)
- [ ] Navigate to other components that used buttons and links (accordion, alert, callout, menus, breadcrumbs, text with image)

### Functional Review Steps
- [ ] Verify everything still looks and functions as expected

### Design Review

### Accessibility Review


kara-franco:
Hiya! The weird Netlify display issue is happening again. 

Screen Shot 2022-04-04 at 10 11 37 AM

ahitson (approved):

ahitson:
I forgot about the underline on hover! Please add an underline to hover and focus.

kara-franco (changes requested):
Looks awesome, excellent work refactoring all that.

- [x] The only item I noticed is the need for underline on hover/focus (like Ashley mentioned).

ModulesUnraveled:
@ahitson @kara-franco I added `text-decoration: underline` when the cta is an `<a>` tag. This is different from the way we do the "slide-in" thing on text links because that would get applied below the entire button box, not just the text inside the button... If this works, great. If not, we (I) may have to figure out a refactor that allows me to add the underline slide in effect inside a button.

With that said, I'm not sure that would be the best idea, since the slide in might clash with the various hover effects that are possible for buttons (fade, slide-up, and wipe)

ahitson:
I love the underline just the way it is. Is the underline black on the blue text on focus, though?

ModulesUnraveled:
@ahitson No, I zoomed in super close, just make sure, and the underline follows the text color

kara-franco (approved):

callinmullaney (approved):
@ModulesUnraveled This PR looks good to me and the refactor here seems like it puts your code in a more organized state. I really dig how you're using components in other files by just referencing the component grouping `@use '../../../02-molecules/molecules';` instead of the individual component itself `@use '../../../02-molecules/menu/menu';`. I need to refactor my project to follow this convention :+1: 

The only comment I really have here is with this refactor the `_button.scss` seems like the forgotten file. It only contains reset styles which makes me think this should be rolled into the new cta scss file as well. Otherwise, if you plan on adding more mixins into this file in the future I could see it being useful.

ModulesUnraveled:
@callinmullaney The plan is to eventually create a bare-bones `<button>` component that can be used for things like the main nav toggle (that should look anything like a CTA) so that they don't have all of the extra data-attribute cruft that is used to achieve the CTA styles. Right now, that toggle (and the accordion/alert toggles, etc) all have that markup, but aren't USING the markup. So we figured we'd eventually create a "basic" `button` component and refactor those to use that instead... but in another ticket. So for this one I just moved that "button reset" into that directory for now.

PR #57, opened 2022-04-04
2) YALB-93: News Card (List and Grid)

ModulesUnraveled:
## This PR addresses the following tickets:
- [YALB-92: News Card (List)](https://yaleits.atlassian.net/browse/YALB-93)
- [YALB-103: News List](https://yaleits.atlassian.net/browse/YALB-93)
- [YALB-93: News Card (Grid)](https://yaleits.atlassian.net/browse/YALB-93)
- [YALB-104: News Grid](https://yaleits.atlassian.net/browse/YALB-93)

### Description of work
- Creates the News Card component with appropriate style options for list and grid
- Creates a "Card Collection" component that can display items as a list or grid

### Testing Link(s)
- [ ] Navigate to the [News Card Story](https://deploy-preview-57--dev-component-library-twig.netlify.app/?path=/story/molecules-cards--news-card)
- [ ] Navigate to the [News Card Story](https://deploy-preview-57--dev-component-library-twig.netlify.app/?path=/story/organisms-card-collection--news-card-collection)

### Functional Review Steps
- [ ] Verify all of the controls work as expected

### Design Review

### Accessibility Review
- [ ] Verify the component meets Accessibility requirements (particularly when the cards are linked!)

### From Kara
- [x] The single news card or list should not be wrapped in a `<li>` OR should have a `<ul>` wrapped around it. Currently, the `<li>` is not wrapped which is throwing an accessibility error. 
- [x] More @ahitson / design related feedback, but the most current edits of the News Cards/Lists were linked (entire card/list). I've linked some edited design files below.

### From Ashley
- [x] Make the whole card clickable (heading black, then blue/underlined on hover. Image drop shadow - level 1)
- [x] Divider between list items (drop-shadow on hover)
- [x] Secondary has date if there is no image or description)
- [x] Update list width to "feature" (grid stays max)


ModulesUnraveled:
@ahitson I definitely want you to review the Percy build for this. A lot of the general link stuff changed (I think for the better?) If you want we can do that together.

kara-franco (changes requested):
Looking great! 

Screen Shot 2022-04-05 at 12 13 46 PM

[New Text List](https://www.figma.com/proto/l84dYRVCvA5DbbCkG0yWke/UI-Kit-%5BYaleSites%5D?node-id=5028%3A48152&scaling=min-zoom&page-id=3296%3A41685&starting-point-node-id=3296%3A41686&show-proto-sidebar=1&hide-ui=1)
[New List Visual](https://www.figma.com/proto/l84dYRVCvA5DbbCkG0yWke/UI-Kit-%5BYaleSites%5D?node-id=4826%3A48454&scaling=min-zoom&page-id=3296%3A41685&starting-point-node-id=3296%3A41686&show-proto-sidebar=1&hide-ui=1)
[News List/Card](https://www.figma.com/proto/l84dYRVCvA5DbbCkG0yWke/UI-Kit-%5BYaleSites%5D?node-id=5041%3A50116&scaling=min-zoom&page-id=3296%3A41685&starting-point-node-id=3296%3A41686&hide-ui=1)

ModulesUnraveled:
@kara-franco I think I've addressed all of your feedback!

@ahitson I've addressed everything except the one left unchecked. This is actually trickier than I though it would be, let's jump on a zoom tomorrow to see what we can figure out.

components/01-atoms/controls/text-link/_text-link.scss
```diff
@@ -58,12 +48,85 @@
...
+  // The `__heading-link` that applies to the appropriate component.
+  $selector: &;
+
+  // Style the component when it's heading is focused.
+  &:focus-within {
+    @include tokens.focus-styles;
+  }
+
+  // Style the heading link inside the component.
+  #{$selector}__heading-link {
```
callinmullaney, line 85: Another valid approach here may be to drop the `$selector` part of this mixin and just use `[class*='__heading-link']`. This might simplify the mixin a bit because you'll be doing less mapping. 

Either way, I think this mixin is understandable.

callinmullaney (changes requested):
@ModulesUnraveled Having the parent component responsible for the design of it's children is a nice touch and definitely simplifies the rendered markup. I do wonder if this approach will cause conflicts if a card listing was embedded into a card grid. However, that might never be a valid scenario.

That being said, a couple of things I want to point out in addition to my code comments.
1. The hover state for multi-line header links does not seem correct to me:
multi-line-links
2. This component is named `News Card` and I wonder if you should use a more generic term. The date field doesn't appear to be a required element so when this is absent it just seems like a generic card component. In theory, this design/layout could be used for any type of content a child site wants to implement.
3. Consider what the Grid/List would look like with mixed content. Some having images and some without. Unless the view the backend dev is planning to build explicitly filters out content with/without images you'll likley see this mix once site's are actually using it.

components/01-atoms/controls/text-link/_text-link.scss
ModulesUnraveled: I like this better. It's still scoped to the component, so there shouldn't be any spillage, and I like not having the interpolated variable in there. 👍

ModulesUnraveled:
@callinmullaney
- I talked with Ashley, and there shouldn't ever be a case where there's a list inside a grid. The example that we thought looked like that will be two components next to each other, not nested. So I think we're okay on that front.
- We noticed the underline issue, yours is actually showing up differently than on my/Ashley's computer (both are wrong, but different). We created a ticket to handle wrapping links because it affects more than just cards. e.g. links in the body that wrap lines.
- I considered making this card generic since the Event Card will share a lot of the same markup and styles, but I think they're different enough to keep them separate. Maybe there will be some "card-base" type of styles or something... we'll figure that out when we do the events.
- The intention is that there will always be an image, or it will be a "default" image. However, I have coded it so that if the image is not provided (for some reason) or if the image fails to load, the "content" part will still line up with the content of the adjacent cards. So I think we're covered here.

I think I've addressed the other code changes

kara-franco (approved):
Very nice, my issues are cleared up. 

I like the idea mentioned in your reply about a "card-base" styles eventually! 

Also, if it helps for your debugging of the wrapped link, this is how it presents on my end: 
Screen Shot 2022-04-08 at 2 59 31 PM

components/02-molecules/cards/news-card/news-card.stories.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Storybook Definition.
+ */
+export default {
+  title: 'Molecules/Cards',
+  argTypes: {
+    date: {
+      name: 'Date',
+      type: 'string',
```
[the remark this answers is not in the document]
callinmullaney: Well that sucks :man_shrugging: oh well.

callinmullaney (approved):
@ModulesUnraveled Code changes look good and all :ok_hand: with your responses. I think the `card_base` may become necessary in a future refactor but we'll cross that bridge when we get there.

:shipit:

ahitson:
Going to approve it since the actual authoring tool will prevent the unwanted combos. BUT, if there is a way to tweak the list toggles, I would really enjoy that. It would be nice if: 
- List, featured (false), images (false), was reflective of the condensed version with just the date and headline.
- and List, featured (false), images (true) had the larger headline and no description.

ahitson (approved):

miketullo95 (approved):

PR #79, opened 2022-04-21
YALB-89: Banner component

ModulesUnraveled:
### Description of work
- Creates the banner component

### Testing Link(s)
- [ ] Navigate to the [Banner Story](https://deploy-preview-79--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--banner)
- [ ] Navigate to the [Standard Page With Banner Example](https://deploy-preview-79--dev-component-library-twig.netlify.app/?path=/story/page-examples-standard-pages--with-banner)

### Functional Review Steps
- [ ] Verify the CTA Banner looks correct (play with all controls at all sizes)

### Design Review

### Accessibility Review


ahitson:
Will there be a text link or CTA option? The current CTA isn't doing the underline behavior.

ModulesUnraveled:
@ahitson I forgot about the link style. I added that support, and fixed the issue of the cta not having an underline.

kara-franco (approved):
Looks great!

ahitson:
Please make this component a top of page item that will visually hide the H1.

ahitson (approved):

components/01-atoms/controls/cta/_cta.scss
```diff
@@ -125,17 +126,20 @@ $cta-outline-weights: (
     color: var(--color-cta-text-active) !important;
     box-shadow: var(--cta-box-shadow-active) !important;
   }
+
+  // @at-root docs: https://sass-lang.com/documentation/at-rules/at-root
+  @at-root #{selector.unify(&, a)} {
```
callinmullaney, line 131: This is awesome. That is all :fist_right: :fist_left:

components/02-molecules/banner/cta/cta-banner.twig
```diff
@@ -0,0 +1,60 @@
...
+  <div {{ bem('content-wrapper', [], banner__base_class) }}>
+    <div {{ bem('image', [], banner__base_class) }}>
+      {% block banner__image %}
+        {% include "@atoms/images/image/_responsive-image.twig" %}
+      {% endblock %}
+    </div>
+    <div {{ bem('content', [], banner__base_class) }} data-component-theme={{ banner__content__background }}>
+      <div {{ bem('text', [], banner__base_class) }}>
+        {% include "@atoms/typography/headings/heading.twig" with {
+          heading__level: '1',
```
callinmullaney, line 31: This is the only line that is sticking out to me as something we might want to change. Currently, this is statically setting the heading to be an `h1` which is what the linked design calls for. However, if this Banner CTA is used or repurposed for any other type of instance (available via a components field, repurposed as a carousel slide) it will cause errors. I wonder if there should be a `banner__heading_level` variable that defaults to `1` but can be easily updated to any other number during implementation.

But it does meet the requirements of the current design so feel free to say "will refactor when appropriate"

components/05-page-examples/_intro-content-examples.twig
```diff
@@ -1,25 +1,41 @@
 {% if intro_content == 'image' %}
   {% include "@molecules/image/content-image.twig" %}
-{% elseif intro_content =='image--highlight' %}
+{% elseif intro_content == 'image--highlight' %}
```
callinmullaney, line 3: I wish Twig would implement a switch case statement. Other CMS's implemented this functionality but it would be a perfect fit for this file. https://www.branchcms.com/learn/docs/developer/twig/tags/item/switch

callinmullaney (approved):
@ModulesUnraveled This PR looks good to me. Marking as approved but consider the `banner__heading_level` comment I left.

components/02-molecules/banner/cta/cta-banner.twig
ModulesUnraveled: Yeah, I originally had it as an h2 actually, but found out that this will REPLACE the page title, if used. So the current expectation is that this IS the page title. There are no current designs to suggest this will be used in a carousel, or anywhere else on the page.

Jim and I discussed the possibility that these could be used in the content area at a later time. But if we do that, there will be other (Drupal) refactoring required anyway, so we'll handle the frontend portion at that time, if the need comes up.

components/05-page-examples/_intro-content-examples.twig
ModulesUnraveled: I agree

PR #90, opened 2022-05-20
YALB-110: Example Events Page

ModulesUnraveled:
### Description of work
- Creates an example events page

### Testing Link(s)
- [ ] Navigate to the [Event Page Example](https://deploy-preview-90--dev-component-library-twig.netlify.app/?path=/story/page-examples-events--event-page)

### Functional Review Steps
- [ ] Verify you can install the components with the CLI

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/file/l84dYRVCvA5DbbCkG0yWke/UI-Kit-%5BYaleSites%5D?node-id=3296%3A41685)

### Accessibility Review


ahitson:
@kara-franco just want to double-check that we are using small caps for event dates as well, or if it is only for news items. Otherwise, it looks good.

kara-franco (changes requested):
Excellent! I like the idea of holding the event meta information in an independent component- my thoughts are that this could help when we want to add teasers to Search results, we can pull in the Event meta and have less work. 

The only accessibility item I found is the markup for the "CTA for Event" vs the "Add to Calendar". I know that the calendar widget isn't built out yet, but my thinking is that we want:  

- [ ] Add to Calendar to be a <button>. 
- [ ] CTA for Event will most likely be a link to register (Eventbrite ex) and should be a <a> element.

 

Screen Shot 2022-05-31 at 4 13 40 PM

ahitson (approved):

miketullo95 (approved):

ModulesUnraveled:
@callinmullaney I abstracted both meta types into a common `meta` folder like we discussed. I think this will work, because we might also use those components on a search results page, or something else that's not a page-title. And since it is doing the "add a pipe between items" thing, it is providing value that the text alone does not.

@kara-franco I agree, I just had the example link in the wrong place, and a typo. The controls should have the correct elements now.

kara-franco (approved):

callinmullaney (approved):
@ModulesUnraveled Good call with abstracting this to a `basic-meta`. I dig how this is organized. :shipit:

PR #92, opened 2022-06-08
YALB-555 and 556: Event item and Collection

ModulesUnraveled:
### Description of work
- Refactors `news-card` to `reference-card` so that it can be used for both news and events

### Testing Link(s)
- [x] Navigate to the [News Card Story](https://deploy-preview-92--dev-component-library-twig.netlify.app/?path=/story/molecules-cards--news-card)
- [x] Navigate to the [Event Card Story](https://deploy-preview-92--dev-component-library-twig.netlify.app/?path=/story/molecules-cards--event-card)
- [ ] Navigate to the [News Collection Story](https://deploy-preview-92--dev-component-library-twig.netlify.app/?path=/story/organisms-card-collection--news-card-collection)
- [ ] Navigate to the [Event Collection Story](https://deploy-preview-92--dev-component-library-twig.netlify.app/?path=/story/organisms-card-collection--event-card-collection)

### Functional Review Steps
- [x] Verify you can install the component with the CLI
- [x] Verify the news and event cards both look and function as expected

### Design Review
- [ ] Verify the News card still looks correct
- [ ] Verify the Event card looks correct
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/file/l84dYRVCvA5DbbCkG0yWke/UI-Kit-%5BYaleSites%5D?node-id=3296%3A41685)

### Accessibility Review


callinmullaney (changes requested):
@ModulesUnraveled Refactor looks pretty good and I like the change to `reference card` :+1: 
1. The event card has a format field that allows me to check both in-person and online. The result of which are `Online,In-person`. Note the no space after the comma. Can these values both be true? If so I think it should read `Online & In-person`. If these things both can't be true then I would suggest changing this control to a select list or radio.
2. Will a card always be an `<li>`? I wonder if this should be converted to `<{{ card__element }}>` and all it's use case to be more flexible when wired to Drupal templates. 
3. As a follow up to the previous feedback I would suggest piping the wrapping `bem` in a `card__attributes` array and using `add_attributes` on the card wrapper itself. If the card element can be a div there might be instances where an attribute would need to be added.

nJim (approved):
I'm glad to see other SMEs chiming in for feedback. From a strict implementation side of things, the code passes all of my smell tests.

ModulesUnraveled:
@ahitson I made the change we talked about and put the format in the overline. You'll also see the time in list mode. (I couldn't figure out how to make it use a different time zone, so UTC it is - for the component library - In Drupal, it'll use the correct time zone.)

@callinmullaney
1. I fixed the format thing. When both are selected, it's supposed to say "Hybrid." I had that working for the event-meta, but forgot to implement it here. 
2. Yes, these cards will always be an `<li>` Even if there's only one "upcoming event", it'll still be in a list markup.
3. I know we've talked about this before, but I don't think we need to add the extra lines and complexity when all we currently support is a simple class name. I'd rather keep it as-is, and refactor if/when it's needed.

kara-franco (changes requested):
Very nice!! I like the use of Heydon's pattern with using `<li>`!

- [x] Only item I caught was the use of the `<h3>` for the date. Can this be updated to regular text? The use of a heading adds clutter for screen reader users, since the dates are not true sub-sections fo the page. 

Screen Shot 2022-06-14 at 5 00 27 PM
Screen Shot 2022-06-15 at 9 41 44 AM

ahitson:
Looking good. This is going to sound ridiculous, but it looks like it is adding a tiny space after the date in the event so the pipe has more space to the left than the right. And is there a way to decrease the height of the pipe a smidge? Maybe one step smaller in font size?
Screen Shot 2022-06-15 at 10 54 04 AM

callinmullaney (approved):
Format change to `Hybrid` looks good on my end. :+1:

ModulesUnraveled:
@kara-franco I changed the subheading element to a simple `<div>` since, like you said, it's not actually a subheading in this context. Let me know if that doesn't work for some reason.

@ahitson I remove the extra space, but the pipe is just a `border-left` so it's exactly the same height as the text - in this case that means it's equal to the line height. There isn't any way to change it's height except to bring down the line height... but I don't think that's a good idea.

ahitson:
Approving this with the caveat that we fully reconsider small caps used on the event cards/lists for time and date. It reads like wild spaghetti language and places too many character types in a small space. We have all caps, sentence case, and small caps which displays a mix of all caps and title case. Both of my UX and UI sides are saying that this is bad.

ahitson (approved):

nJim:
@ModulesUnraveled Do you have changes that have not been pushed up? You mentioned some changes in this conversation but I do not see any changes on this branch. This would explain why the H3 is still there in netlify.

ModulesUnraveled:
Blerg, and Ugh! Yeah, I had done everything locally, and not pushed anything up. It should be there now. @kara-franco @nJim

ModulesUnraveled:
@kara-franco I need this in a release in order to wrap up my wiring work. So I'm going to merge, but if you (or anyone else) have additional feedback, we can get it addressed in a follow-up ticket.

PR #99, opened 2022-06-30
YALB-561: text with image spacing

codeb-a:
### Description of work
- updates the spacing for text with image molecule.

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-99--dev-component-library-twig.netlify.app/?path=/story/tokens-breakpoints--breakpoints)

### DIFF
--- a/components/02-molecules/text-with-image/_text-with-image.scss
+++ b/components/02-molecules/text-with-image/_text-with-image.scss
@@ -8,12 +8,12 @@
   display: grid;
 
   @media (max-width: tokens.$break-mobile-max) {
-    gap: var(--size-spacing-5);
+    gap: var(--size-spacing-6);
   }
 
   @media (min-width: tokens.$break-mobile) {
     grid-template: 'primary secondary';
-    gap: var(--size-spacing-6);
+    gap: var(--size-spacing-8);
 
     // Focus Options
     [data-component-focus='equal'] & {


components/02-molecules/text-with-image/_text-with-image.scss
```diff
@@ -8,12 +8,12 @@
...
 
   @media (max-width: tokens.$break-mobile-max) {
-    gap: var(--size-spacing-5);
+    gap: var(--size-spacing-6);
   }
 
   @media (min-width: tokens.$break-mobile) {
     grid-template: 'primary secondary';
-    gap: var(--size-spacing-6);
+    gap: var(--size-spacing-8);
```
ModulesUnraveled, line 16: @ahitson I just wanted to point out that we have a couple of "global" gutter sizes:

```
  --spacing-component-gutter: var(--size-spacing-9);
  --spacing-component-gutter-secondary: var(--size-spacing-7);
```

For example, the card grid uses those (`-gutter` for the featured card grid and `-gutter-secondary` for the non-featured grid) and the two-column layout uses `-gutter` for the sidebar.

Did you want to use one of those for this as well? If so, this should change from 8 to 7 or 9. Not sure if making this "line up" with other components is important to you, or if it's desirable to be 8 specifically.

Just wanted to throw that out.

ModulesUnraveled: @ahitson Like we discussed on zoom. We'll keep this as a custom value for now, and can re-evaluate how components share gutter values later. 👍

ModulesUnraveled (approved):
Code's good. As long as @ahitson approves the design side, you can :shipit:

ahitson:
There is some funky stuff going on with the link, but I have a feeling that is due to the underline work so I'll ignore it. Yeah?

ahitson (approved):

ModulesUnraveled:
Oh, yeah, that does have to do with the link stuff... It should look right when all of that gets merged.

PR #100, opened 2022-06-30
YALB-597: Fix link underline animation

ModulesUnraveled:
### Description of work
- Refactors the link underline thing to support multiple lines
- Implements the new animation speed tokens to keep things consistent

### Testing Link(s)
- [ ] Navigate to the [News Cards Story](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/organisms-card-collection--news-card-collection)

### Design Review
- [ ] Verify the heading link underline thing animates multiple lines

### DIFF
--- a/components/00-tokens/effects/_effects.scss
+++ b/components/00-tokens/effects/_effects.scss
@@ -1,6 +1,6 @@
 @mixin animate(
   $property: all,
-  $duration: 200ms,
+  $duration: var(--animation-speed-default),
   $function: ease-in-out,
   $delay: 0ms
 ) {
--- a/components/01-atoms/controls/text-link/_text-link.scss
+++ b/components/01-atoms/controls/text-link/_text-link.scss
@@ -26,39 +26,27 @@
 // Style options are "underline" or "no-underline".
 // The style name indicates the initial appearance.
 @mixin link($style: underline, $family-weight: 'mallory-medium') {
-  @include tokens.animate('color');
-
-  position: relative;
   color: var(--link-color-base);
   text-decoration: none;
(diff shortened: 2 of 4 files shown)


callinmullaney (changes requested):
@ModulesUnraveled Both my local and deploy-preview are showing the background gradient style blocking out the entire text. Looking into this further it doesn't appear that `var(--link-styles-no-underline-size)` is being set.

Also, I think `background-size` might need two values in order to get working fully. When I added `background-size: 0 2px` for the default and ``background-size: 100% 2px` for the hover I was getting something closer to what I think was intended.

ModulesUnraveled:
@callinmullaney Can you verify you have the latest tokens update on your local (it's in the updated lock file). That should have the code you need to see it not completely blocked out.

@kara-franco Can you re-run the build without cache, or whatever so that Netlify picks up the updated tokens as well?

callinmullaney (approved):
@ModulesUnraveled Updating my local token definitions seems to have fixed things for me and the deploy preview looks good as well. :shipit:

ahitson:
Can we get the underline closer to the baseline?

kara-franco (approved):
Looks great!

ModulesUnraveled:
@ahitson, as we figured on Zoom, it looks like setting the line position to `1.15em` will likely work for these. However... since I've already started working on the tabs component which uses this underline thing, and refactored the underlying code in that work, will you be okay if I address the underline position as a part of that PR? That will also give us another use-case to make sure that that value will work across the various typography styles.

miketullo95 (approved):
looks good!

PR #102, opened 2022-07-02
YALB-78: Pagination

ModulesUnraveled:
### Description of work
- Creates the initial Pagination component
- Creates Event Grid page (should have had this before)
- Creates Event List page including the pager at the bottom

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-102--dev-component-library-twig.netlify.app/?path=/story/molecules-navigation-pager--basic)

### Functional Review Steps

### Design Review

### Accessibility Review

### DIFF
--- a/components/02-molecules/pager/pager.twig
+++ b/components/02-molecules/pager/pager.twig
@@ -1,0 +1,59 @@
+{% set pager__base_class = 'pager' %}
+
+{% if items %}
+  <nav {{ bem(pager__base_class) }} role="navigation" aria-label="{{ heading_id }}">
+    <ul {{ bem('items', [], pager__base_class, ['js-pager__items']) }}>
+      {# Print previous item if we are not on the first page. #}
+      {% if items.previous %}
+        <li {{ bem('item', ['prev'], pager__base_class) }}>
+          <a {{ bem('link', ['prev'], pager__base_class) }} href="{{ items.previous.href }}" title="{{ 'Go to previous page'|t }}" rel="prev"{{ items.previous.attributes|without('href', 'title', 'rel') }}>
+            <span {{ bem('visually-hidden') }}>{{ 'Previous page'|t }}</span>
+            <span aria-hidden="true">{{ items.previous.text|default('‹ Previous'|t) }}</span>
+          </a>
+        </li>
+      {% endif %}
+      {# Print first item if we are not on the first page. #}
+      {% if items.first %}
+        <li {{ bem('item', ['first'], pager__base_class) }}>
+          <a {{ bem('link', ['first'], pager__base_class) }} href="{{ items.first.href }}" title="{{ 'Go to first page'|t }}"{{ items.first.attributes|without('href', 'title') }}>
+            <span {{ bem('visually-hidden') }}>{{ 'First page'|t }}</span>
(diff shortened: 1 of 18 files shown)


ahitson:
- The active page should skip hover/focus.
- Change the color of the outline and active page number to basic-brown-gray.
- Can we increase the padding between previous/last and the page numbers?
- Is there a way to see mobile?

ModulesUnraveled:
@ahitson I've addressed all of your requests!

ahitson:
- Change "previous" to "back"
- Use the chevron icon instead of the text arrows < >
- On the basic 3/5 mobile there aren't directional controls the directional controls should appear

ModulesUnraveled:
@ahitson Done

ahitson:
- Add size spacing 1 margin between the back/next and icon
- Change the icon size to .8 em
- Make directional arrows always present except when on the 1st (no back) and last pages (no next)

ahitson (approved):

kara-franco (changes requested):
Excellent!! Three requests to make for a better screen reader user experience: 

- [ ] Add `aria-label="Pagination"`to the `<nav>`
Screen Shot 2022-07-11 at 12 48 00 PM

- [ ] The title attribute + the link name are really redundant for screen reader user, they hear "Link, Page 1, Go to page 1". Refer to the US Web Design System for arial-label example: https://designsystem.digital.gov/components/pagination/ 

Screen Shot 2022-07-11 at 1 23 35 PM

- [ ] Add `aria-current="page" `to current page, so users know where they are. 

Screen Shot 2022-07-11 at 12 57 22 PM

miketullo95 (approved):

components/02-molecules/pager/pager.twig
```diff
@@ -0,0 +1,79 @@
+{% set pager__base_class = 'pager' %}
```
callinmullaney, line 1: This file seems like it should be split into partial at the very least (see main feedback from the PR review). It's using 4 different types of markup with at least 4 levels of embedding said markup. If you decide to not extend the base menu for pager I would probably partial out this file to `pager-list.twig`, `pager-item.twig`, and `pager-link.twig`.

ModulesUnraveled: I think I disagree... this is a very different component from a typical "menu". There are a lot of custom things this needs to support (`items.previous` (and next), `items.first` (and last), and then the "regular" `items.pages` - They're all mixed together in one array - and no other menu has those things.

This one also needs to pass icons into the content of specific items, which I think would be a pain to shoehorn into the base menu component.

There's also the fact that the "desktop" and "mobile" pagers are both here, but each are hidden at relevant screen sizes.

Also, the base menu components add a lot of extra stuff like menu levels - which will never apply to a pager.

On a side - but related - note, I wanted to talk with you about our implementation of the list component. I'm not sure the value is worth the complexity it adds. Let's zoom tomorrow, if you're free.

ModulesUnraveled:
@callinmullaney I'm going to merge this to unblock a release. I scheduled a zoom for us to go over your feedback, and if anything needs to be done as a result, I'll open a followup PR 👍

PR #112, opened 2022-07-28
YALB-474: Controls refactor

ModulesUnraveled:
### Description of work
- Refactors all previous uses of the cta component that should not have had all of the cta markup

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-112--dev-component-library-twig.netlify.app/?path=/story/tokens-breakpoints--breakpoints)

### Functional Review Steps
- [ ] Verify everything still looks and functions as expected.

PR #113, opened 2022-07-28
YALB-350: Component tokens

ModulesUnraveled:
### Description of work
- Refactors component tokens out of the component library

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-113--dev-component-library-twig.netlify.app/?path=/story/tokens-breakpoints--breakpoints)

### Functional Review Steps
- [ ] Verify everything still looks and functions as expected. You can probably do your functional checking in https://github.com/yalesites-org/component-library-twig/pull/112 since that includes all of the refactor stuff


callinmullaney (approved):
@ModulesUnraveled This looks pretty good from what I can tell. I see we're having formatting conflicts between prettier and stylelint in `_config.scss`. Instead of disabling stylelint do you think it worth doing different comment formatting? Maybe just documentation about `::root {` instead of comment blocks inside? Either way, I don't really see that as being a blocker.

ModulesUnraveled:
@callinmullaney That was actually just a personal preference. I wanted to keep all of the color stuff together - rather than having a bunch of empty lines between each of them. This file is a little "special" because it'll be used to indicate **what** can be set at a theme level. I think most of the lines in here will eventually all be commented out by default, and simply there as indications of what can be set.

PR #121, opened 2022-08-10
YALB-627: hiding breadcrumbs short trails

codeb-a:
### Description of work
- Will not display breadcrumbs if there are only 2 segments. Only if 3 or more levels deep.

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-121--dev-component-library-twig.netlify.app/?path=/story/tokens-breakpoints--breakpoints)

### Functional Review Steps

### Design Review

### Accessibility Review

### DIFF
--- a/components/03-organisms/menu/breadcrumbs/breadcrumbs.twig
+++ b/components/03-organisms/menu/breadcrumbs/breadcrumbs.twig
@@ -12,6 +12,7 @@
 {% set parent_count = breadcrumbs__items.length - 2 %}
 {% set parent_label = breadcrumbs__items[parent_count].title %}
 
+{% if breadcrumbs__items.2 %}
 <div {{ add_attributes(breadcrumbs__attributes) }}>
   {% set control__content %}
     <span {{ bem('visually-hidden') }}>Show all breadcrumbs</span>
@@ -37,3 +38,4 @@
     } %}
   </div>
 </div>
+{% endif %}


components/03-organisms/menu/breadcrumbs/breadcrumbs.twig
```diff
@@ -12,6 +12,7 @@
 {% set parent_count = breadcrumbs__items.length - 2 %}
 {% set parent_label = breadcrumbs__items[parent_count].title %}
 
+{% if breadcrumbs__items.2 %}
```
ModulesUnraveled, line 15: I would normally put this here as well. But I was just thinking that this means lines 1-13 are still technically "running" even if they're not used. If we move the `if` to the very top, then none of that logic will run.

I'm not actually sure if this is the way Twig works though... if this was JS for the browser, it definitely would be, but I think Twig renders on the server side, so it might not matter...

I asked in the frontend channel in Slack. We'll see if anyone else knows the answer.

ModulesUnraveled: Let's keep it here. Sounds like there isn't really a benefit since Twig (php) will render server side anyway. This is definitely easier to read 👍

ModulesUnraveled (approved):
Looks good to me! :shipit:

PR #126, opened 2022-08-17
YALB-650: update date and time for news/event pages

codeb-a:
### Description of work
- updates the date and time of longform articles and events to small-caps and oldstyle numerals.

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-126--dev-component-library-twig.netlify.app/?path=/story/tokens-breakpoints--breakpoints)

### Functional Review Steps

### Design Review

### Accessibility Review


components/02-molecules/page-title/page-title.stories.js
```diff
@@ -16,7 +16,7 @@ export default {
       type: 'string',
       defaultValue: `<span>By Charlyn Paradis</span>${dateTimeTwig({
         date_time__start: '2022-01-25',
-        date_time__format: 'date',
```
ModulesUnraveled, line 19: I'm not sure why you changed these... In the designs it looks like it's using the `date` format (without the day of the week) to me. However, I think the event meta looks like it should be using `day__full`, so that could be changed in `yds-event-meta.twig`.

You can see the various date formats in `yds-date-time.twig` starting on line 19, if you're curious what they are.

components/02-molecules/meta/event-meta/_yds-event-meta.scss
```diff
@@ -9,6 +9,12 @@
 
 .event-meta__date {
   @include tokens.body-xl;
+
+  .date-time {
```
ModulesUnraveled, line 13: Two things jumped out to me in this file:
1. There is a class nested inside another class. That's generally a no-no. Sometimes, it's unavoidable, but should definitely set off alarm bells when you see it.
2. You're writing this code twice (in this file, and the `_yds-basic-meta.scss` file. Twice is just a warning, if I saw it in three places, that would be an "error" in my books.

And luckily, both of those concerns will be easy to fix here 😄 

Looking at this image of the inspector, you'll see that the `font-variant: small-caps` was already being applied to both of these, and in both cases, it's coming from the same file `_yds-date-time.scss`. So, I think you should be able to move this css all into that one file, and both instances (as well as future date implementations) will get the new styles.
Screen Shot 2022-08-19 at 2 52 51 PM
 
Give that a shot, and let me know if you run into any issues.

ModulesUnraveled (changes requested):

components/02-molecules/meta/basic-meta/_yds-basic-meta.scss
```diff
@@ -20,4 +20,10 @@
       border-left: var(--border-thickness-1) solid;
     }
   }
+
+  .date-time {
+    font-variant: small-caps;
+    font-variant-numeric: oldstyle-nums;
+    text-transform: lowercase;
+  }
```
ModulesUnraveled, line 28: This can be removed, now that you added these styles to the `.date-time` element in the other file. If you inspect your browser, you'll see that this is overriding those styles, with exactly the same styles. So it's not providing any benefit here.

ModulesUnraveled (changes requested):
Just one last change, and then this should be good! 👍

ahitson:
Text does look better, but it's creating some up and down numbers that are very woah-no for me. Any way to keep those from rollercoastering?

Also, the dividing pipe in the byline and date/time seems like it's growing in length, especially on the list views. Is there another solution we can use to split that content? A • maybe? Anything?


Screen Shot 2022-08-23 at 5 06 27 PM

ModulesUnraveled:
@ahitson that's the `oldstyle-numbs` part... and OPAC specifically asked for it.

I agree that it looks really bad, in my opinion... but I'm not sure what to say about it as it's specifically requested in the ticket: https://yaleits.atlassian.net/browse/YALB-650

ModulesUnraveled:
And the pipe is just a `border-left` on the text, so it's exactly the same height as the line-height. But now that we're tweaking the text (to make it smaller) but not affecting the line-height, it is standing out more...

components/02-molecules/meta/basic-meta/_yds-basic-meta.scss
```diff
@@ -17,7 +17,12 @@
     > *:not(:first-child) {
       margin-left: var(--size-spacing-5);
       padding-left: var(--size-spacing-5);
-      border-left: var(--border-thickness-1) solid;
+    }
+
+    .date-time::before {
+      position: relative;
+      content: '|';
+      right: var(--size-spacing-5);
```
ModulesUnraveled, line 25: Similar to above, you have this only applying to the `data-time` element in the meta, and we actually want it to apply to _anything_ in the meta, so this can be moved up inside the previous `> *:not(:first-child)` section.

Also, the positioning isn't quite right. If you inspect it, you'll see it's off center. There are a few things we need to do to make it exactly centered.

- [ ] Let's add `display: flex` to the `basic-meta` class, so that there's no extra space between these inline elements.
- [ ] Set `position: relative` on the items (above the `margin-left` and `padding-left` stuff)
- [ ] Set `position: absolute` on the `before` element
- [ ] You'll notice the `right` positioning you have here isn't actually doing anything in the browser (there's a technical limitation of custom properties that makes them not available to pseudo-elements like `::before`), and since this is content it's getting a little extra space on the left, so let's used a specific value - in this case I think `left: -0.225rem;` will make it align the way we want.

ModulesUnraveled (changes requested):
Looking good. I've made some suggestions. Just make sure all of the suggestions here apply to both implementations

components/02-molecules/cards/reference-card/_yds-reference-card.scss
```diff
@@ -79,7 +79,12 @@
   > *:not(:first-child) {
     margin-left: var(--size-spacing-5);
     padding-left: var(--size-spacing-5);
-    border-left: var(--border-thickness-1) solid;
+
+    &::before {
+      position: relative;
+      content: '|';
+      right: var(--size-spacing-5);
+    }
```
ModulesUnraveled, line 87: This one still has the wrong `position` and `right` values. Please add `position: relative` above, and change this to `absolute` and the change `right` to `left` with the value to match the other one.

ModulesUnraveled (changes requested):

ModulesUnraveled:
@codeb-a per the recent comments in the Jira ticket, please remove the old-style numerals code

components/01-atoms/date-time/_yds-date-time.scss
```diff
@@ -1,3 +1,5 @@
 .date-time {
   font-variant: small-caps;
+  font-variant-numeric: oldstyle-nums;
+  text-transform: lowercase;
```
ModulesUnraveled, line 4: Per the recent Jira comments, let's remove the oldstyle-nums code here.

ModulesUnraveled (approved):
This looks good to me!

PR #134, opened 2022-09-01
YALB-293: Namespace event format

marcb4k:
### Description of work
- Renames event format twig file to comply with namespaces
- Fixes event cards and lists templates from Drupal

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-134--dev-component-library-twig.netlify.app/?path=/story/organisms-card-collection--event-card-collection)
- [ ] Verify that event collections still work

### Functional Review Steps
None

### Design Review
None

### Accessibility Review
None


ModulesUnraveled (approved):
Everything works fine. Did you find you needed this in Drupal? I was thinking we wouldn't need to namespace the partials (like `_event-format`, but not for any specific reason. If you do need to call this directly from a Drupal template though, then I definitely think we _should_ namespace them.

Everything seems to work fine, in Storybook, so as far I'm concerned, we can shipit.

marcb4k:
@ModulesUnraveled - This was needed in Atomic, so it was sort of a "do we fix here or in Atomic" question. No problem, glad it's in there now. Thanks!

PR #145, opened 2022-09-16
YALB-711: creates video component

codeb-a:
### Description of work
- Creates video atom and text-with-video molecule.

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-145--dev-component-library-twig.netlify.app/?path=/story/tokens-breakpoints--breakpoints)

### Functional Review Steps

### Design Review

### Accessibility Review


components/01-atoms/video/video.twig
```diff
@@ -0,0 +1,23 @@
...
+ * - video_modifiers - array of modifiers to add to the base classname
+ *   by default, videos are expected to be 16x9, include the modifier "full" to support a 4x3 video
+ * - video_blockname - blockname prepended to the base classname
+ * - video_content - the content of the video (typically an iframe)
+ *   the "|raw" filter is applied so that the iframe is rendered instead
+ *   of simply passing the iframe as plain text.
+ *
+ * Available blocks:
+ * - video_content - used to replace the content of the video with something other than the typical iframe
+ *   for example: to insert an html5 video component
```
ModulesUnraveled, line 14: Please update the variables to use the double underscore method of separating the variable parts. So, change `video_content` to `video__content`.

Also, you can remove all of the BEM-related ones from the docs here. I figure that those are a given, and so to keep this section most helpful, we can remove the `base_class`, `modifiers`, and `blockname` variables from the docs... because they'd basically be the same thing in every file.

ModulesUnraveled (changes requested):
I think what you have as `text-with-video` is going to be more like the `content-image` than the `text-with-image`, so let's meet on Monday to talk that through.

Everything's looking good, there's just a few tweaks I think we need to make. And I think this will be a great contribution back to Compound once we finish the refactor!

components/02-molecules/video/video.twig
```diff
@@ -0,0 +1,39 @@
+{#
+ # Available variables:
+ # - text_with_video__heading
+ # - text_with_video__text
+ #
+ # Available blocks:
+ # - text_with_video__video
+ #}
```
ModulesUnraveled, line 8: We forgot to update the variable names in this file. They should all be something like `video__base_class` instead of `text_with_video__base_class`

components/02-molecules/video/video.yml
```diff
@@ -0,0 +1,3 @@
+text_with_video__heading: 'This Is Where A Video Title Will Go'
+text_with_video__text: '<p>This is where captions for videos will go.</p>'
```
ModulesUnraveled, line 2: Make sure you update these two variables to match the changes in the twig.

As a matter of fact, you can probably do a find and replace over your codebase for `text_with_video__` and replace it with just `video__`

Make sure to look through the files it's going to change, but I think it should work.

components/05-page-examples/standard-pages/standard-page.stories.js
```diff
@@ -279,3 +281,51 @@ WithQuickLinks.argTypes = {
...
+withVideo.argTypes = {
+  videoHeading: {
+    name: 'Video Heading',
+    type: 'string',
+    defaultValue: videoData.text_with_video__heading,
+  },
+  videoCaption: {
+    name: 'Video',
+    type: 'string',
+    defaultValue: videoData.text_with_video__text,
```
ModulesUnraveled, line 329: Just a note about the variable names here

ModulesUnraveled (changes requested):
We just forgot to refactor one file, but I think this will be good to go after that!

ModulesUnraveled (approved):
Code looks good now! Thanks for all your work!

ahitson:
- Change the title color to gray 800 and the caption to gray 700. 
- The title should use Base sizing and the caption is Small.

ModulesUnraveled:
@codeb-a per Ashley's feedback, there should be a variable for the `heading` color, and another called something like `text-muted` or something. You should be able to find them in the image component's css

ntay90 (approved):

components/02-molecules/video/_yds-video.scss
```diff
@@ -0,0 +1,22 @@
...
+  display: grid;
+  gap: var(--size-spacing-3);
+}
+
+.video__heading {
+  margin-bottom: var(--size-spacing-1);
+}
+
+.video__text {
+  font: var(--font-style-body-s);
```
ModulesUnraveled, line 17: Sorry! I should have caught this before... all typography should use one of the mixins we have. So, in this case, instead of 
```
font: var(--font-style-body-s);
```
you should do
```
@include tokens.body-s;
```

You can see all of the possible options when you toggle the "Font details" under one of the body styles here: https://deploy-preview-145--dev-component-library-twig.netlify.app/?path=/story/tokens-typography--body-styles (The heading ones have it too)

This is because some of the styles (really just the headings at this point, but we want to keep implementation consistent) have things like letter-spacing or text-transforms applied, and that can't be captured as a part of a single css custom property

ModulesUnraveled (changes requested):
Sorry!!!! One last code change

ahitson:
One last thing, please up the margin-bottom on the heading to spacing-2.

ahitson (approved):

ModulesUnraveled (approved):
:shipit:

PR #146, opened 2022-09-21
YALB-707: Image grid

ModulesUnraveled:
### Description of work
- Creates the Image Grid component

### Testing Link(s)
- [x] Navigate to the [Component Story](https://deploy-preview-146--dev-component-library-twig.netlify.app/?path=/story/organisms-galleries--image-grid)

### Functional Review Steps
- [x] Verify you can install the component with the CLI

### Design Review

### Accessibility Review


ntay90 (approved):

codeb-a (approved):

ahitson:
Breakpoints! We need to add some here....will send something to you in slack.

components/03-organisms/galleries/media-grid/yds-media-grid.twig
```diff
@@ -0,0 +1,34 @@
...
+ #}
+
+{% set media_grid__base_class = 'media-grid' %}
+
+{% set media_grid__attributes = {
+  'data-component-width': 'feature',
+  'data-media-grid-variation': media_grid__variation|default('basic'),
+} %}
+
+<div {{ bem(media_grid__base_class) }} {{ add_attributes(media_grid__attributes) }}>
```
callinmullaney, line 17: `bem()` should be moved up into `media_grid__attributes`

components/03-organisms/galleries/media-grid/yds-media-grid.twig
```diff
@@ -0,0 +1,34 @@
...
+  <div {{ bem('inner', [], media_grid__base_class) }}>
+    {% if media_grid__heading %}
+      {% include "@atoms/typography/headings/yds-heading.twig" with {
+        heading__level: '2',
+        heading__blockname: media_grid__base_class,
+        heading: media_grid__heading,
+      } %}
+    {% endif %}
+    <ul {{ bem('items', [], media_grid__base_class) }}>
```
callinmullaney, line 26: This seems like it should be embedding the generic list component instead of using the `<ul>` directly. Is your thinking here that doing so just complicates this component?

callinmullaney (changes requested):
@ModulesUnraveled Left a couple code comments but for the most part things look good. I'm curious about why you chose to organize this component under `organisms` and not `molecules`? Other than the headings atom this component is rather flat/simple and (in my head) doesn't fit within the concept of an organism.

components/03-organisms/galleries/media-grid/yds-media-grid.twig
ModulesUnraveled: Done

components/03-organisms/galleries/media-grid/yds-media-grid.twig
ModulesUnraveled: Our other collections are currently written this way. I don't see any value add at the moment for the added complexity that would come with implementing the list component.

If we did, this:
```
{% set media_grid__base_class = 'media-grid' %}

{% set media_grid__attributes = {
  'data-component-width': 'feature',
  'data-media-grid-variation': media_grid__variation|default('basic'),
  'class': bem(media_grid__base_class),
} %}

<div {{ add_attributes(media_grid__attributes) }}>
  <div {{ bem('inner', [], media_grid__base_class) }}>
    {% if media_grid__heading %}
      {% include "@atoms/typography/headings/yds-heading.twig" with {
        heading__level: '2',
        heading__blockname: media_grid__base_class,
        heading: media_grid__heading,
      } %}
    {% endif %}
    <ul {{ bem('items', [], media_grid__base_class) }}>
      {% block media_grid__items %}
        {% for item in media_grid__items %}
          {% include "@organisms/galleries/media-grid/_yds-media-grid-item.twig" %}
        {% endfor %}
      {% endblock %}
    </ul>
  </div>
</div>
```

would become:
```
{% set media_grid__base_class = 'media-grid' %}

{% set media_grid__attributes = {
  'data-component-width': 'feature',
  'data-media-grid-variation': media_grid__variation|default('basic'),
  'class': bem(media_grid__base_class),
} %}

{% set media_grid__items %}
  {% block media_grid__items %}
    {% for item in media_grid__items %}
      {% include "@organisms/galleries/media-grid/_yds-media-grid-item.twig" %}
    {% endfor %}
  {% endblock %}
{% endset %}

<div {{ add_attributes(media_grid__attributes) }}>
  <div {{ bem('inner', [], media_grid__base_class) }}>
    {% if media_grid__heading %}
      {% include "@atoms/typography/headings/yds-heading.twig" with {
        heading__level: '2',
        heading__blockname: media_grid__base_class,
        heading: media_grid__heading,
      } %}
    {% endif %}
    {% embed "@atoms/lists/yds-list.twig" with {
      list__base_class: 'items',
      list__blockname: media_grid__base_class,
    } %}
      {% block list__content %}
        {{ media_grid__items }}
      {% endblock %}
    {% endembed %}
  </div>
</div>
```

Adding the extra "intermediate block" is a sort of anti-pattern. It works, but it's not real clean, and can make things more confusing. i'm not going to do the refactor now. But if we ever come across a reason to have a shared "list" component across the three collections (card-collection, custom-card-collection, and media-grid) then we can refactor them together.

ModulesUnraveled:
@callinmullaney I replied to your code comments inline, and as far as this being an organism - the reason is that this is the "basic" version. There's going to be an "interactive" one (where images pop up in a modal) and possibly variations that are carousel-like. So, while this one is super simple (it's just images) the variations will be more complex... so it's equivalent (in my head) to the card collections.

@ahitson I finally got the breakpoints in, and while I was at it, consolidated how the card grids are done so that everything works more-or-less the same. Please review all three "Card Collections" and the "Image Grid" all using both the "featured" and not options, at all breakpoints, to make sure I didn't inadvertently break something!

components/03-organisms/galleries/media-grid/yds-media-grid.twig
callinmullaney: That makes sense to me. Although, I do think we should strive to reuse things as much as possible I see what you're saying about it be a bit of an anti-pattern. I think we may need to fundamentally rethink the list component in general so it's easier to re-use and doesn't require this added external block. But this is future-Brian/Callin's problem. 

Approving :+1:

callinmullaney (approved):
:shipit:

ModulesUnraveled:
@ahitson The grid has been updated. Please take another look!

components/03-organisms/galleries/media-grid/_yds-media-grid.scss
```diff
@@ -0,0 +1,59 @@
...
+  @include tokens.h2-yale-new;
+
+  margin-bottom: var(--spacing-page-inner);
+}
+
+.media-grid__items {
+  @include grid.secondary(true);
+}
+
+[data-media-grid-variation='interactive'] {
```
callinmullaney, line 18: To follow convention we established in other components I think this should be wrapped inside `.media-grid__item` with a parent selector? Or was this authored in this manner to avoid grid within a grid inheritance?

components/03-organisms/galleries/media-grid/_yds-media-grid.scss
```diff
@@ -0,0 +1,59 @@
...
+    opacity: 0.8;
+  }
+}
+
+.media-grid__maximize {
+  position: absolute;
+  right: var(--size-spacing-4);
+  bottom: var(--size-spacing-4);
+  background: var(--color-basic-white);
+  height: 1.5rem;
```
callinmullaney, line 43: These height/width styles correspond with `var(--size-spacing-6)` and `var(--size-spacing-4)`. It makes sense to not apply them here as these variables aren't necessarily for dimensions. However, I think these would be better handled be a generic `var(--icon-size-height)` and `var(--icon-size-width)` or something more generic like `var(--icon-size-small)` - It looks like you've already implemented something similar that is specific for Alerts. Consider consolidating these styles: https://github.com/yalesites-org/component-library-twig/blob/develop/components/02-molecules/alert/_yds-alert.scss#L9

components/03-organisms/galleries/media-grid/yds-media-grid-interactive.js
```diff
@@ -0,0 +1,42 @@
...
+      items.forEach((item) => {
+        item.addEventListener('click', () => {
+          grid.setAttribute(
+            'data-media-grid-modal-state',
+            toggleModalState(modalState),
+          );
+        });
+      });
+    });
+
```
callinmullaney, line 26: Consider adding a keyboard trap to this file. I noticed I'm immediately able to tab-outside of the modal - https://github.com/emulsify-ds/compound/pull/35/files#diff-266cf28551632cfda8819f7b4b2b2a449d4fe849742c74188265dab8205c81bfR26

components/03-organisms/_grid-mixins.scss
```diff
@@ -0,0 +1,58 @@
...
+      flex: 0 1 calc(50% - calc(var(--grid-gutter) / 2));
+    }
+
+    @media (min-width: tokens.$break-l) {
+      flex: 0 1 calc(33.33% - calc(var(--grid-gutter) * 2 / 3));
+    }
+  }
+}
+
+@mixin secondary($media-only: false) {
```
callinmullaney, line 33: Logic here looks good and can be merged but I would consider making this more generic if you are inclined. Instead of passing "what" the grid is going to be used for you could pass the breakpoint where you want 2 columns to kick-in. So `@include grid.secondary(true);` could be written as `@include grid.secondary('break-s');` <-- I think this is easier to discern what the argument passing is doing at implementation. 

soooo 

```
@mixin secondary($breakpoint: null) {
...
> * {
    // Media-only grids are always two-up
    @if $breakpoint {
      // Grids with content (like the news cards) start one-up, and go to two-up
      // at this breakpoint.
      @media (min-width: tokens.$breakpoint) {
        flex: 0 1 calc(50% - calc(var(--grid-gutter) / 2));
      }
    } @else {
        flex: 0 1 calc(50% - calc(var(--grid-gutter) / 2));
    }
   ...
```

callinmullaney (changes requested):

ahitson (approved):

components/03-organisms/galleries/media-grid/_yds-media-grid.scss
ModulesUnraveled: Good call. I think I this happened after a refactor, and I just missed it.

components/03-organisms/galleries/media-grid/_yds-media-grid.scss
ModulesUnraveled: These particular heights and widths are actually for the wrapping "circle" around the icon. But I did update the icon itself to follow that pattern.

It would be good to see if we can identify a set of sizes for icons across the board and reuse them... but I don't know if that's a thing. Either way, I've updated this implementation to be consistent with the one in alerts.

Thanks!

components/03-organisms/galleries/media-grid/yds-media-grid-interactive.js
ModulesUnraveled: Fo sho... all of this interactive stuff is actually coming in a follow-up PR (already heavily underway) I think I just accidentally committed some of the initial work to this branch instead of that one... Oops!

I'll add the keyboard trap in the next PR.

components/03-organisms/_grid-mixins.scss
ModulesUnraveled: I disagree here. While it would make it more flexible, the reason the news grids don't start at two-up is specifically because they have text in addition to the image. And the reason the "image-only" grids are always two-up - even at the smallest screen sizes - is specifically because they don't have any "extra information".

So, to allow someone to place news cards in a grid that start at two-up would break design requirements because the text area would be so narrow, you'd only get one word on a line (or they might even overlap the next card). We want to actively avoid that, and just giving a wide-open, use-case-less (if that's a word) variable makes that mistake easier to make.

So, I think the logic (and variable naming) for this use-case is correct as is:

> If there's no additional text to display with the image - it's always at least two-up

> If there is supplemental text shown below the image - be one-up until the browser is wide enough to support showing that textual content in a two-up display

components/03-organisms/galleries/media-grid/_yds-media-grid.scss
callinmullaney: ah yes you're right regarding the circle wrapper. Lets revisit the icon size in a follow up PR if this type of implementation needs to carry through to other components. This at least looks consistent with the alerts implementation.

components/03-organisms/_grid-mixins.scss
callinmullaney: I see what you're saying. So is 
```
.media-grid__items {
  @include grid.secondary(true);
}
``` 
essentially the only instance this will be used given other grid items will contain text? For me, what prompted this feedback was the single use case for this mixin argument and the fact that the implementation `@include grid.secondary(true);` isn't clear what it's doing. This mixin is a bit on-the-rails to enforce design requirements which I think is fine but wanted to be clear on your thought process around it.

callinmullaney (approved):
@ModulesUnraveled I left another comment regarding `grid.secondary` but this is good to merge. Nice work!

components/03-organisms/_grid-mixins.scss
ModulesUnraveled: I was on the fence about how to use the mixin. Since it's the first (and only) argument, we don't have to name it when we use it. But I went ahead and named it in the usage so that it's more clear as to what the argument is for. I'm going to merge this now, but if you think this could still be improved, let's do it in the "interactive grid" PR.

Here's what it looks like now, if you don't want to review the code:
```
.media-grid__items {
  @include grid.secondary($media-only: true);
}
```

PR #147, opened 2022-09-22
YALB-700: Search Form

marcb4k:
### Description of work
- Adds a search form to the utility navigation area
- NOTE: This requires [PR-115](https://github.com/yalesites-org/yalesites-project/pull/115) from yalesites-project and [PR-46](https://github.com/yalesites-org/atomic/pull/46) from atomic

### Functional testing steps:
- [ ] Visit any page and verify that the search box is located in the utility menu area and is styled appropriately (See screenshot)

Screen Shot 2022-09-22 at 10 56 36 AM

### DIFF
--- a/components/03-organisms/menu/utility-nav/yds-utility-nav.twig
+++ b/components/03-organisms/menu/utility-nav/yds-utility-nav.twig
@@ -2,21 +2,22 @@
 
 <div {{ bem('utility-bar') }}>
   {# Menu #}
-  {% include "@molecules/menu/yds-menu.twig" with {
-    menu__base_class: utility_nav__base_class,
-    menu__name: 'Utility',
-    items: utility_nav__items|default(items),
-  } %}
+  {% block utility_nav__menu %}
+    {% include "@organisms/menu/utility-nav/_utility-nav--menu.twig" %}
+  {% endblock %}
   {# Search #}
+  {% set utility_nav__search = true %}
   {% if utility_nav__search %}
     <div {{ bem('search', [], utility_nav__base_class) }}>
-      {% include "@atoms/images/icons/_yds-icon.twig" with {
-        icon__name: 'magnifying-glass',
-        icon__blockname: utility_nav__base_class,
-        icon__modifiers: ['search'],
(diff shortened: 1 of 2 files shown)


components/03-organisms/menu/utility-nav/yds-utility-nav.twig
```diff
@@ -2,21 +2,22 @@
...
-  {% include "@molecules/menu/yds-menu.twig" with {
-    menu__base_class: utility_nav__base_class,
-    menu__name: 'Utility',
-    items: utility_nav__items|default(items),
-  } %}
+  {% block utility_nav__menu %}
+    {% include "@organisms/menu/utility-nav/_utility-nav--menu.twig" %}
+  {% endblock %}
   {# Search #}
+  {% set utility_nav__search = true %}
```
ModulesUnraveled, line 9: We can't do this here, or it will force search to be on for every site. The `if` below is there so that sites can choose whether or not they want to have a site search.

Many of the sites on the platform will be essentially "brochureware sites" with only a handful of pages, and search would seem ridiculous. So we want to provide a toggle in the Drupal Settings page for them to turn search on or off.

ModulesUnraveled (changes requested):
Need to remove the hardcoded variable.

To test, you should be able to go to the Basic Page (http://localhost:6006/?path=/story/page-examples-standard-pages--basic) and toggle the `Header: Search` control to see it appear and dissapear.

components/03-organisms/menu/utility-nav/yds-utility-nav.twig
marcb4k: Completely understand - this is removed in YALB-702 and turned into a dynamic variable that comes from the theme settings form. This was hardcoded here so that we could test YALB-700.

ModulesUnraveled (approved):
Knowing that a follow-up PR already exists to fix this, I'm approving it. Thanks!

ModulesUnraveled:
I'm going to merge this so that #702 is a clean diff

PR #158, opened 2022-11-02
 YALB-795: Callout edits in Drupal

joetower:
### Description of work
- fix: p:last-child margin removal. 
- added an `if` statement around callout content so the markup doesn't render if no content has been entered

### Testing Link(s)

### Functional Review Steps

### Design Review

### Accessibility Review

### DIFF
--- a/components/02-molecules/callout/_yds-callout.twig
+++ b/components/02-molecules/callout/_yds-callout.twig
@@ -14,10 +14,12 @@
       heading__blockname: callout__base_class,
       heading: callout__heading,
     } %}
-    {% include "@atoms/typography/text/yds-text.twig" with {
-      text__blockname: callout__base_class,
-      text__content: callout__text,
-    } %}
+    {% if callout__text %}
+      {% include "@atoms/typography/text/yds-text.twig" with {
+        text__blockname: callout__base_class,
+        text__content: callout__text,
+      } %}
+    {% endif %}
     {% if callout__link__url and callout__link__content %}
       {% if callout__link__type == 'link' %}
         {% include "@atoms/controls/text-link/yds-text-link.twig" with {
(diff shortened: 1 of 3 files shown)


components/02-molecules/callout/callout.yml
```diff
@@ -1,5 +1,5 @@
 callout__heading: 'Degree Programs'
-callout__text: 'Designed for those who intend to pursue graduate and those who wish to immediately enter a career in which broad scientific training is beneficial.'
+callout__text: '<p>Designed for those who intend to pursue graduate and those who wish to immediately enter a career in which broad scientific training is beneficial.</p><p>Designed for those who intend to pursue graduate and those who wish to immediately enter a career in which broad scientific training is beneficial.</p>'
```
ModulesUnraveled, line 2: I don't think Ashley will want this demoed with two paragraphs. It's good to see it working that way, but let's remove it from the Storybook example so that people don't think that they're supposed to do that.

components/02-molecules/callout/_yds-callout.twig
```diff
@@ -14,10 +14,12 @@
       heading__blockname: callout__base_class,
       heading: callout__heading,
     } %}
-    {% include "@atoms/typography/text/yds-text.twig" with {
-      text__blockname: callout__base_class,
-      text__content: callout__text,
-    } %}
+    {% if callout__text %}
```
ModulesUnraveled, line 17: This works in Storybook, but doesn't actually fix it in Drupal.

Since the `paragraph--callout-item.html.twig` file is just passing `callout__text: content.field_text` the text wrapper is still showing (because of Drupal's render array thing. Even though there's no content in the field, there are still items in the array - so it evaluates to true)

You'll need to update that to `callout__text: content.field_text[0]` to check whether there's actually content in the field. Then, this should work.

ModulesUnraveled (changes requested):
Exactly what I was thinking. Just need to remove the extra paragraph from Storybook, and update the Drupal template.

components/02-molecules/callout/_yds-callout.twig
joetower: @ModulesUnraveled I have added a branch to fix this here: https://github.com/yalesites-org/atomic/pull/60

Thank you!

components/02-molecules/callout/callout.yml
joetower: @ModulesUnraveled I removed the extra `p` element

ModulesUnraveled (approved):
Thanks for wrapping this up! :shipit:

PR #161, opened 2022-11-03
YALB-696: Header navigation alignment

AnaBZC:
### Description of work
- fix: primary-nav__link--level-0:last-child padding right removal

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-161--dev-component-library-twig.netlify.app/?path=/story/organisms-site-header--header)

### Functional Review Steps
- [ ] Verify that the header navigation is aligned
- [ ] Verify that the `primary-nav__link--level-0:last-child` does not have padding

### DIFF
--- a/components/03-organisms/menu/primary-nav/_yds-primary-nav.scss
+++ b/components/03-organisms/menu/primary-nav/_yds-primary-nav.scss
@@ -214,3 +214,9 @@
     transform: rotate(180deg);
   }
 }
+
+.primary-nav__link--level-0:last-child {
+  @media (min-width: tokens.$break-mobile) {
+    padding-right: 0;
+  }
+}
--- a/components/03-organisms/menu/primary-nav/_yds-primary-nav.scss
+++ b/components/03-organisms/menu/primary-nav/_yds-primary-nav.scss
@@ -44,6 +44,10 @@
       .primary-nav__item--level-0:first-child & {
         padding-left: var(--size-spacing-5);
       }
+
+      .primary-nav__item--level-0:last-child & {
+        padding-right: var(--size-spacing-5);
+      }
     }
 
     .primary-nav__item--level-0:first-child & {
--- a/components/03-organisms/menu/primary-nav/_yds-primary-nav.scss
+++ b/components/03-organisms/menu/primary-nav/_yds-primary-nav.scss
@@ -41,6 +41,14 @@
       border-color: var(--color-navigation-border);
       border-bottom-color: var(--color-background);
     }
+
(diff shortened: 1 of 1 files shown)


components/03-organisms/menu/primary-nav/_yds-primary-nav.scss
```diff
@@ -214,3 +214,9 @@
     transform: rotate(180deg);
   }
 }
+
+.primary-nav__link--level-0:last-child {
```
ModulesUnraveled, line 218: This doesn't work because it's affecting every `.primary-nav__link--level-0` since they're all the "last-child" of their respective `<li>` elements. (see screenshot) We'll need to target the last `.primary-nav__item--level-0` and then apply these styles to the link inside that.

Screen Shot 2022-11-03 at 4 35 34 PM

It doesn't affect anything that has the class `.primary-nav__toggle--level-0` which is everything that has child links, so it's not super obvious.

Also, we need to remove the `padding-left` from the first item, to make it align as well.

There is a mixin at the top of that file `@mixin primary-nav-item-level-0` that is implemented in both `.primary-nav__link--level-0` and `.primary-nav__toggle-level-0` where shared styles should go. I think that's going to be the right place to put this css.

Inside the media query on line 31, I think I'd add this, which should make it work the way we expect.

```
.primary-nav__item--level-0:first-child & {
  padding-left: 0;
}

.primary-nav__item--level-0:last-child & {
  padding-right: 0;
}
```

ModulesUnraveled (changes requested):

components/03-organisms/menu/primary-nav/_yds-primary-nav.scss
```diff
@@ -40,6 +40,18 @@
       z-index: 1;
       border-color: var(--color-navigation-border);
       border-bottom-color: var(--color-background);
+
+      .primary-nav__item--level-0:first-child & {
+        padding-left: var(--size-spacing-5);
+      }
```
ModulesUnraveled, line 46: Good call on this! It's too bad that it makes it "jump" when you expand and collapse it... but I don't know what we can do to avoid it.

We should also add the padding-right to the last item, in case it's a toggle, instead of a link.

ModulesUnraveled (changes requested):

ModulesUnraveled (approved):
Perfect. Thank!

ahitson:
Link to the component story is broken. Where do I look?

ahitson (approved):

PR #166, opened 2022-11-04
YALB-510: Alerts: Fade away animation

joetower:
### Description of work
- Adds functionality bullet item
- Fixes this or that bullet item

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-166--dev-component-library-twig.netlify.app/?path=/story/molecules-alert--alert-examples)
- [ ] Dismiss an alert and verify it now fades out

### Functional Review Steps

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/proto/G4sJ0BLu4jCJpGzW82JHE8/YaleSites-Components?node-id=4769%3A8322&scaling=min-zoom&page-id=4769%3A7045)

### Accessibility Review

### DIFF
--- a/components/02-molecules/alert/_yds-alert.scss
+++ b/components/02-molecules/alert/_yds-alert.scss
@@ -31,6 +31,8 @@ $alert-fade-speed: var(--animation-speed-slow);
   }
 
   &[data-alert-state='dismissed'] {
+    @include tokens.animate-hidden($alert-fade-speed);
+
     opacity: 0;
     max-height: 0;
     padding: 0;
--- a/components/02-molecules/alert/_yds-alert.scss
+++ b/components/02-molecules/alert/_yds-alert.scss
@@ -17,6 +17,8 @@ $alert-fade-speed: var(--animation-speed-slow);
   color: var(--color-alert-text);
   opacity: 1;
 
+  @include tokens.animate(opacity, $alert-fade-speed);
+
   @each $theme, $value in $alert-themes {
     &[data-alert-type='#{$theme}'] {
       --color-alert-background: var(--color-alert-#{$theme}-background-color);
@@ -26,8 +28,6 @@ $alert-fade-speed: var(--animation-speed-slow);
   }
 
   &[data-alert-state='dismissed'] {
-    @include tokens.animate(opacity, $alert-fade-speed);
-
     opacity: 0;
   }
 }
@@ -40,13 +40,13 @@ $alert-fade-speed: var(--animation-speed-slow);
   max-height: 7rem;
   opacity: 1;
 
+  @include tokens.animate(all, $duration: $alert-fade-speed);
+
   @media (min-width: tokens.$break-mobile) {
     align-items: center;
   }
 
   [data-alert-state='dismissed'] & {
-    @include tokens.animate(all, $alert-fade-speed);
-
     opacity: 0;
     max-height: 0;
     padding: 0;
--- a/components/02-molecules/alert/_yds-alert.scss
+++ b/components/02-molecules/alert/_yds-alert.scss
@@ -5,6 +5,7 @@
 
 $alert-themes: map.deep-get(sass-tokens.$tokens, color, alert);
 $alert-fade-speed: var(--animation-speed-slow);
+$alert-inner-height-speed: var(--animation-speed-default);
 
 .alert {
   --size-icon: 1.25rem;
@@ -26,8 +27,7 @@ $alert-fade-speed: var(--animation-speed-slow);
   }
 
   &[data-alert-state='dismissed'] {
-    @include tokens.animate(opacity);
-    @include tokens.animate-hidden($alert-fade-speed);
+    @include tokens.animate(opacity, $alert-fade-speed);
 
     opacity: 0;
   }
@@ -37,9 +37,20 @@ $alert-fade-speed: var(--animation-speed-slow);
   display: flex;
   align-items: flex-start;
   padding-block: var(--size-spacing-5);
+  height: 100%;
+  max-height: 7rem;
+  opacity: 1;
 
   @media (min-width: tokens.$break-mobile) {
     align-items: center;
+  }
+
+  [data-alert-state='dismissed'] & {
+    @include tokens.animate(all, $alert-inner-height-speed);
+
+    opacity: 0;
+    max-height: 0;
+    padding: 0;
   }
 }
 
@@ -55,10 +66,7 @@ $alert-fade-speed: var(--animation-speed-slow);
   @media (min-width: tokens.$break-mobile) {
     margin-right: var(--size-spacing-5);
 
-    [data-alert-state='expanded'] & {
-      align-self: flex-start;
-    }
-
+    [data-alert-state='expanded'] &,
     // fix icon alignment shift
     [data-alert-state='dismissed'] & {
       align-self: flex-start;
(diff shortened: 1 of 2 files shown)


components/02-molecules/alert/_yds-alert.scss
```diff
@@ -4,6 +4,7 @@
 @use '../../01-atoms/atoms';
 
 $alert-themes: map.deep-get(sass-tokens.$tokens, color, alert);
+$alert-fade-speed: var(--animation-speed-slow);
 
 .alert {
   --size-icon: 1.25rem;
```
ModulesUnraveled, line 10: Right after these custom properties are set, we can use the `animate` mixin to make our "visual" changes. I like how you did opacity, but I'm also going to animate the `max-height` to prevent the little bit of "overlapping text" that can happen the way it is now. So I'd just add this:

```
  @include tokens.animate($duration: $alert-fade-speed);
```

And then after the `color` definition, let's default alerts to

```
opacity: 1;
max-height: 7rem; // This is just a random number I chose. If we go too much higher, it'll make the animation speed seem fast... and it should be plenty large enough to accommodate the amount of text they want to allow in the alerts.
```

components/02-molecules/alert/_yds-alert.scss
```diff
@@ -24,7 +25,7 @@ $alert-themes: map.deep-get(sass-tokens.$tokens, color, alert);
   }
 
   &[data-alert-state='dismissed'] {
-    display: none;
+    @include tokens.animate-fade-to-hidden($alert-fade-speed);
```
ModulesUnraveled: Here, we can use the animate-hidden mixin with the speed variable we set at the top of this file.

```
@include tokens.animate-hidden($alert-fade-speed);
```

Then, we need to set the new "visual" attributes. I'm thinking something like:

```
    opacity: 0;
    max-height: 0;
    overflow: hidden;
```

components/02-molecules/alert/_yds-alert.scss
```diff
@@ -53,6 +54,11 @@ $alert-themes: map.deep-get(sass-tokens.$tokens, color, alert);
     [data-alert-state='expanded'] & {
       align-self: flex-start;
     }
+
+    // fix icon alignment shift
+    [data-alert-state='dismissed'] & {
+      align-self: flex-start;
+    }
```
ModulesUnraveled, line 65: I was curious what these were for, but I played with it, and I see now. I would probably just lump this in with the chunk above, instead of having the two separate declarations. Because really, we want the "expanded" and the "dismissed" states to be the same size - always, and forever - don't ever let them be different. The way you have it now, someone could tweak one size, and not realize they need to change the other. So we can "couple" them by writing it like this instead:

```
[data-alert-state='expanded'] &,
    [data-alert-state='dismissed'] & {
      align-self: flex-start;
    }
```

And then do the same for the icon size.

ModulesUnraveled (changes requested):
Looking good! Just a couple of tweaks

components/02-molecules/alert/_yds-alert.scss
```diff
@@ -24,7 +26,10 @@ $alert-themes: map.deep-get(sass-tokens.$tokens, color, alert);
   }
 
   &[data-alert-state='dismissed'] {
-    display: none;
+    @include tokens.animate(opacity);
```
ModulesUnraveled, line 29: I think this one has to go on the "default" state? I could be wrong though.

ahitson:
Can we slow down the disappearing banner so that it's a gentle transition out?

ahitson:
Into it.

components/02-molecules/alert/_yds-alert.scss
```diff
@@ -24,18 +26,31 @@ $alert-themes: map.deep-get(sass-tokens.$tokens, color, alert);
   }
 
   &[data-alert-state='dismissed'] {
-    display: none;
+    @include tokens.animate(opacity, $alert-fade-speed);
```
ModulesUnraveled, line 29: This seems to work, but I'm curious why you added it to the dismissed state, rather than to the `.alert` directly

ModulesUnraveled: Oh! I think this isn't actually doing anything, because you're doing the animations to the "inner" element. See below about that.

components/02-molecules/alert/_yds-alert.scss
```diff
@@ -24,18 +26,31 @@ $alert-themes: map.deep-get(sass-tokens.$tokens, color, alert);
...
     align-items: center;
   }
+
+  [data-alert-state='dismissed'] & {
+    @include tokens.animate(all, $alert-fade-speed);
+
+    opacity: 0;
+    max-height: 0;
+    padding: 0;
+  }
```
ModulesUnraveled, line 53: This seems to work, but I'm curious why you added this all to the "inner" element, instead of to the main `.alert` element. I think it'd have added it all to the outer-most element.

Also, the `animate` mixin uses named arguments, so if you just want to affect the speed, you can do that like this
```
@include tokens.animate($duration: $alert-fade-speed);
```

components/02-molecules/alert/_yds-alert.scss
```diff
@@ -14,6 +15,7 @@ $alert-themes: map.deep-get(sass-tokens.$tokens, color, alert);
   z-index: 3;
   background-color: var(--color-alert-background);
   color: var(--color-alert-text);
+  opacity: 1;
```
ModulesUnraveled, line 18: Looks like you're animating the opacity of the outer-wrapper, and the "inner" element. I think we can consolidate that to just the outer-wrapper.

ModulesUnraveled (changes requested):
I think the animation stuff should apply to the outer-most wrapper element. There's some duplication at the moment, and I think it'll clean things up a bit if we keep it all there.

components/02-molecules/alert/_yds-alert.scss
joetower: @ModulesUnraveled :thinking: It was working for me without using `$duration:`. I have added it.

components/02-molecules/alert/_yds-alert.scss
joetower: @ModulesUnraveled This is because I want to animate both the wrapper (`opacity`) with the bg-color and the inner text content (`height`). I want the inner content's height to animate out more gracefully than if we had animations only on the wrapper.

Are you saying that I should apply ` @include tokens.animate(all, $duration: $alert-fade-speed);` to the `.alert` element and remove the `inner` animations? If we do that, the alerts simply collapse and the animation is not noticeable. I'm totally open to whatever suggestions you have, but I'm not clear on how we can simplify this while retaining the animation currently in place. Please let me know where you see opportunities for improvement. 

I did update where the animation was being applied - move to `alert` and `alert__inner` instead of `[data-alert-state='dismissed']`.

components/02-molecules/alert/_yds-alert.scss
```diff
@@ -24,7 +31,9 @@ $alert-themes: map.deep-get(sass-tokens.$tokens, color, alert);
   }
 
   &[data-alert-state='dismissed'] {
-    display: none;
+    opacity: 0;
+    max-height: 0;
+    padding: 0;
```
ModulesUnraveled, line 36: Dang! I forgot one last thing. We're visually hiding this, but one other thing `display: none` was doing was "hiding" it from accessible tech, like screen readers. Right now, the alert would still be read to the user, and stuff. So we need to add the `animate-hidden` mixin to remove it from assistive tech when it is visually hidden.

Adding `@include tokens.animate($alert-fade-speed);` inside the "dismissed" state should do it.

ModulesUnraveled: Also, I don't think we need this `padding: 0` ... at least, I couldn't figure out what it was doing.

ModulesUnraveled: Oh! I bet that was a carry-over from when we had these styles on the `inner` element. I think it can be safely removed now.

joetower: @ModulesUnraveled I pushed up a couple commits addressing these. :crossed_fingers: Should be good once it's done running tests.

ModulesUnraveled (approved):

PR #174, opened 2022-11-11
YALB-833: Grand Hero component

AnaBZC:
### Description of work
- Added Grand Hero component folder to banner folder.
- Added grand hero twig, scss and yml.

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-174--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--grand-hero-banner)

### Functional Review Steps
- [ ] Verify you can select/change content overlay
- [ ] Verify that you can change the overlay of the content to full or contained

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/file/G4sJ0BLu4jCJpGzW82JHE8/YaleSites-Components?node-id=6634%3A22464)

### Accessibility Review


components/02-molecules/banner/grand-hero/_yds-grand-hero.scss
```diff
@@ -0,0 +1,78 @@
...
+.grand-hero-banner__heading {
+  @include tokens.h2-yale-new;
+
+  margin-bottom: var(--size-spacing-2);
+}
+
+.grand-hero-banner__link {
+  @include atoms.link;
+
+  flex: 0 0 auto;
```
ModulesUnraveled, line 77: I couldn't figure out what this was doing. Is it needed?

components/02-molecules/banner/grand-hero/_yds-grand-hero.scss
```diff
@@ -0,0 +1,78 @@
...
+    height: 100%;
+    display: flex;
+    align-items: center;
+  }
+}
+
+.grand-hero-banner__heading {
+  @include tokens.h2-yale-new;
+
+  margin-bottom: var(--size-spacing-2);
```
ModulesUnraveled, line 71: The spacing is off between the text elements. In the designs, it looks like there's a consistent spacing of 9px between the heading, the text, and the link. Since we don't have a 9px variable, I'd suggest we use `--size-spacing-3` since it's equivalent to 8px.

We'll also need to remove the bottom margin from the last `p` tag inside the "snippet" element so that it doesn't create too much space. I'd suggest doing something like we did on the Callout: https://github.com/yalesites-org/component-library-twig/blob/develop/components/02-molecules/callout/_yds-callout.scss#L83-L89

And, since all of the content items are optional, I think using flex or grid (with the `gap` attribute) would be a better solution, so that we don't have dangling margins above, or below any of the items. In order for that to work, you'll need to delete the `__text` element so that everything's at the same level. But I think that's okay - I couldn't see any reason to keep that wrapper.

components/02-molecules/banner/grand-hero/yds-grand-hero.twig
```diff
@@ -0,0 +1,58 @@
...
+ # - grand_hero__snippet
+ # - grand_hero__link__content
+ # - grand_hero__link__url
+ #
+ # Available Blocks:
+ # - grand_hero__media
+ #}
+
+{% set grand_hero__base_class = 'grand-hero-banner' %}
+{% set grand_hero__content__background = grand_hero__content__background|default('basic-black') %}
```
ModulesUnraveled, line 17: `basic-black` isn't one of our color schemes. If you go to the color pairings page (http://localhost:6006/?path=/story/tokens-colors--color-pairings) the darkest one we have is `gray-800`. You might ask Ashley if that will work. Otherwise, we'll need to add a new color scheme.

Actually... I seem to remember her mentioning that this color might be a "custom" color chosen by the theme... if that's the case, we'll need a little more details on what colors can be chosen etc, because we might not even use our `component-theme` thing for this...

components/02-molecules/banner/grand-hero/yds-grand-hero.twig
```diff
@@ -0,0 +1,58 @@
...
+<div {{ add_attributes(grand_hero__attributes) }}>
+  <div {{ bem('content-wrapper', [], grand_hero__base_class) }}>
+    <div {{ bem('image', [], grand_hero__base_class) }}>
+      {% block grand_hero__image %}
+        {% include "@atoms/images/image/_responsive-image.twig" %}
+      {% endblock %}
+    </div>
+    <div {{ bem('content', [], grand_hero__base_class) }} data-component-theme={{ grand_hero__content__background }}>
+      <div {{ bem('content-inner', [], grand_hero__base_class) }}>
+        <div {{ bem('text', [], grand_hero__base_class) }}>
```
ModulesUnraveled, line 34: I _think_ that both of these wrappers can go away, if you make the css changes I mentioned above.

components/02-molecules/banner/grand-hero/yds-grand-hero.twig
```diff
@@ -0,0 +1,58 @@
...
+          {% include "@atoms/typography/headings/yds-heading.twig" with {
+            heading__level: '1',
```
ModulesUnraveled, line 36: This can't be an `h1`. It won't be the page title. I think that it can always be an `h2` though, since it will always be toward the top of the page, and won't be allowed lower in the content.

ModulesUnraveled: I'm not sure about the accessibility of it though... since the actual `h1` of the page will come after this banner in the markup... My quick test shows that it'll be okay... but we'll have their accessibility team check it to be sure.

ModulesUnraveled: Oh, and I just realized that you might have gotten this pattern from the action banner... which also uses an `h1`... that was originally going to _replace_ the page title, when it was used. But they decided they wanted to have the page title be shown _in addition to_ the banner. So, would you please also change the action banner heading to be an `h2` as a part of this PR?

components/02-molecules/banner/grand-hero/yds-grand-hero.twig
```diff
@@ -0,0 +1,58 @@
...
+  <div {{ bem('content-wrapper', [], grand_hero__base_class) }}>
+    <div {{ bem('image', [], grand_hero__base_class) }}>
+      {% block grand_hero__image %}
+        {% include "@atoms/images/image/_responsive-image.twig" %}
+      {% endblock %}
+    </div>
+    <div {{ bem('content', [], grand_hero__base_class) }} data-component-theme={{ grand_hero__content__background }}>
+      <div {{ bem('content-inner', [], grand_hero__base_class) }}>
+        <div {{ bem('text', [], grand_hero__base_class) }}>
+          {% include "@atoms/typography/headings/yds-heading.twig" with {
```
ModulesUnraveled, line 35: This is optional, so we need to wrap it in an `if`

components/02-molecules/banner/grand-hero/yds-grand-hero.twig
ModulesUnraveled, line 34: Since all of the fields are optional, we should wrap the content wrapper in an `if` that checks to make sure at least one of them is provided so that we don't print empty divs.

components/02-molecules/banner/grand-hero/yds-grand-hero.twig
```diff
@@ -0,0 +1,58 @@
...
+          {% include "@atoms/controls/text-link/yds-text-link.twig" with {
+            link__content: grand_hero__link__content,
+            link__url: grand_hero__link__url,
+            link__blockname: grand_hero__base_class,
+          } %}
+        {% endif %}
+      </div>
+    </div>
+  </div>
+</div>
```
ModulesUnraveled, line 58: Need a new line at the end of the file.

ahitson:
On the full overlay, the text should use the same width rules as the contained version so that it doesn't go across the screen.

Screen Shot 2022-11-16 at 10 30 39 AM

components/02-molecules/banner/grand-hero/yds-grand-hero.js
```diff
@@ -0,0 +1,8 @@
+Drupal.behaviors.grandHero = {
+  attach() {
+    const header = document.querySelector('.site-header');
+    document
+      .getElementsByClassName('grand-hero-banner')[0]
+      .style.setProperty('--site-header-height', header.offsetHeight);
+  },
+};
```
ModulesUnraveled, line 8: This JS didn't work for me. I had to update it to this in order to get it working. You can test it on http://localhost:6006/?path=/story/page-examples-standard-pages--with-banner&globals=outline:true

```
Drupal.behaviors.grandHero = {
  attach() {
    const header = document.querySelector('.site-header');

    window.addEventListener('load', () => {
      const grandHeroes = document.querySelectorAll('.grand-hero-banner');

      grandHeroes.forEach((hero) => {
        hero.style.setProperty(
          '--site-header-height',
          `${header.offsetHeight}px`,
        );
      });
    });
  },
};
```

ModulesUnraveled: However, I don't think this is the right way to do it. I didn't finish my thought yesterday, so I'll try and write it out here:

- [ ] Since we're setting a variable called `--site-header-height` I think that should happen as a part of the `site-header` component. So, I think we can move this JS file into the `site-header` directory. (and rename the behavior to match.)

- [ ] Then, we should set the initial value of the variable in the `04-page-layouts/page-layouts.scss` file. something like:
```
body {
  --site-header-height: 4.8125rem; // 77px / 16
}
```
This way, any future component that needs to react based on the header height also has access to the variable. If it's inside the `.grand-hero-banner` scope, then no other component can access the value, and we'll have to do more js in order for them to have the same value. It'd be better to do it at a high level - since we're doing it anyway, let's just do it once so that this component, and any future ones can all use the same value.

- [ ] Then, we can just set the variable at the "body" scope in JS like this:
```
Drupal.behaviors.siteHeader = {
  attach(context) {
    const body = context.querySelector('body');
    const header = context.querySelector('.site-header');

    window.addEventListener('load', () => {
      body.style.setProperty(
        '--site-header-height',
        `${header.offsetHeight}px`,
      );
    });
  },
};
```

ahitson:
Can you make the padding on the full overlay even on both sides at the med to xsmall breakpoints?
Screen Shot 2022-11-17 at 5 37 46 PM

Also, the underline on the link is hitting the text's baseline. And beige should be removed from the color options.

components/02-molecules/banner/grand-hero/yds-grand-hero.js
ModulesUnraveled: Oh, also, this doesn't adjust when the window resizes. Any time we scale the browser, we should recalculate it. e.g. when the menu wraps to a new line, or on mobile.

I messed with this a little tonight, and got close, but it's not quite working... This should get you most of the way there thought!

```
Drupal.behaviors.siteHeader = {
  attach(context) {
    const body = context.querySelector('body');
    const header = context.querySelector('.site-header');

    /**
     * debounce
     * @description Debounce to only run a function at most once every 200ms.
     * @param {} func The function to be run after the timeout.
     */
    const debounce = (func) => {
      let timer;
      return function debounceFunction(event) {
        if (timer) clearTimeout(timer);
        timer = setTimeout(func, 200, event);
      };
    };

    /**
     * setHeaderHeight
     * @description Set the `--site-header-height` variable.
     */
    const setHeaderHeight = () => {
      console.log(`${header.offsetHeight}px`);
      body.style.setProperty(
        '--site-header-height',
        `${header.offsetHeight}px`,
      );
    };

    // Determine, and set the site header height variable on page load.
    window.addEventListener('load', () => {
      setHeaderHeight();
    });

    // Update the site header height variable when the window is resized.
    window.addEventListener(
      'resize',
      debounce(function runSetHeight() {
        setHeaderHeight();
      }),
    );
  },
};
```

components/02-molecules/banner/grand-hero/_yds-grand-hero.scss
```diff
@@ -0,0 +1,108 @@
...
+  position: relative;
+  display: grid;
+  gap: var(--size-spacing-3);
+
+  @media (min-width: $break-grand-hero-banner) {
+    max-width: var(--size-component-layout-width-content);
+  }
+
+  [data-grand-hero-overlay-variation='full'] & {
+    padding-left: var(--size-spacing-8);
```
ModulesUnraveled, line 90: This should be `padding-left: var(--size-spacing-site-gutter);` because the "site gutter" is the exact variable that needs to be added. (And it changes at breakpoints, so we need to use it so that it's in sync with other components across all screen sizes.)

ModulesUnraveled (changes requested):
JS needs some work, and I left a minor comment on the CSS. This is SUPER close!

ahitson:
Close! It looks like the text padding on the "content overlay: full" version is still not even at the medium breakpoint.

Screen Shot 2022-11-23 at 9 40 08 AM

ahitson:
Padding looks good now. 

On the light background colors, the body text and link should be the same darkest gray color as the heading.

ahitson (approved):

nJim:
Work was approved so I am moving this forward. We need to release related changed. There may be VisReg baselines to update as a result.

PR #176, opened 2022-11-17
YALB-96/600: Search Theming

marcb4k:
### Description of work
- Theming for search page and search results
- Adds components for search results and component wrapper (for wrapping things like views, title, etc)

### Testing Link(s)
- [ ] Navigate to the [Search Result](https://deploy-preview-176--dev-component-library-twig.netlify.app/?path=/story/molecules-search-result--search-result)
- [ ] Navigate to the [Component Wrapper](https://deploy-preview-176--dev-component-library-twig.netlify.app/?path=/story/organisms-component-wrapper--component-wrapper&args=componentWidth:highlight)

### Functional Review Steps

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/proto/G4sJ0BLu4jCJpGzW82JHE8/YaleSites-Components?node-id=6288%3A19237&scaling=min-zoom&page-id=6288%3A19236&hide-ui=1)

### Accessibility Review

### DIFF
--- a/components/02-molecules/search-result/_yds-search-result.scss
+++ b/components/02-molecules/search-result/_yds-search-result.scss
@@ -18,15 +18,38 @@
   padding: var(--size-spacing-5) 0;
   border-bottom: 1px solid var(--color-gray-300);
 
-  .search-result__highlighted,
-  .search-result__teaser {
-    margin: var(--size-spacing-5) 0;
-  }
-
-  .search-result__teaser {
-    color: var(--color-gray-600);
-  }
   @media (min-width: tokens.$break-mobile-max) {
     max-width: var(--size-component-layout-width-content);
   }
+
+  &:hover {
+    // @TODO: Temporary styles until we create a token
+    // This shadow is hard coded because it has to be adjusted to make it
+    // appear as though it's a shadow on only the border-bottom element.
+    // There isn't currently anything else that looks like this. If there ever
+    // Is, perhaps it should be captured as a token in a new drop-shadow set.
+    // @LINK: https://yaleits.atlassian.net/browse/YALB-844.
+    box-shadow: 0 8px 6px -6px hsl(0deg 0% 0% / 16%);
+  }
+}
+
+.search-result__heading {
+  @include tokens.h4-yale-new;
+
+  color: var(--color-gray-800);
+}
+
+.search-result__highlighted,
+.search-result__teaser {
+  margin: var(--size-spacing-5) 0;
+}
+
+.search-result__highlighted {
+  strong {
+    color: var(--color-gray-800);
+  }
+}
+
+.search-result__teaser {
+  color: var(--color-gray-600);
 }
--- a/components/02-molecules/search-result/_yds-search-result.scss
+++ b/components/02-molecules/search-result/_yds-search-result.scss
@@ -1,19 +1,10 @@
+@use '../../01-atoms/atoms';
 @use '../../00-tokens/tokens';
 
 .search-result {
+  @include atoms.clickable-component-heading-link;
+
   border-bottom: 1px solid var(--color-gray-300);
-
-  .breadcrumbs__wrapper {
-    &[data-component-width]:not([data-embedded-components]
-        [data-component-width]) {
-      padding-inline: 0;
-      margin-bottom: 0;
-    }
-
-    [class*='__inner'] {
-      margin-inline: 0;
-    }
-  }
 
   .search-result__highlighted,
   .search-result__teaser {
@@ -24,6 +15,6 @@
     color: var(--color-gray-600);
   }
   @media (min-width: tokens.$break-mobile-max) {
-    max-width: 50%;
+    max-width: var(--size-component-layout-width-content);
   }
 }
--- a/components/02-molecules/search-result/_yds-search-result.scss
+++ b/components/02-molecules/search-result/_yds-search-result.scss
@@ -1,6 +1,8 @@
 @use '../../00-tokens/tokens';
 
 .search-result {
+  border-bottom: 1px solid var(--color-gray-300);
+
   .breadcrumbs__wrapper {
     &[data-component-width]:not([data-embedded-components]
         [data-component-width]) {
@@ -11,6 +13,15 @@
     [class*='__inner'] {
       margin-inline: 0;
     }
+  }
+
+  .search-result__highlighted,
+  .search-result__teaser {
+    margin: var(--size-spacing-5) 0;
+  }
+
+  .search-result__teaser {
+    color: var(--color-gray-600);
   }
   @media (min-width: tokens.$break-mobile-max) {
     max-width: 50%;
--- a/components/02-molecules/search-result/_yds-search-result.scss
+++ b/components/02-molecules/search-result/_yds-search-result.scss
@@ -1,0 +1,3 @@
+.search-result .breadcrumbs__wrapper {
+  padding-inline: 0;
+}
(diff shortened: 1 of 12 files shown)


components/02-molecules/search-result/_yds-search-result.scss
```diff
@@ -0,0 +1,32 @@
...
+  border-bottom: 1px solid var(--color-gray-300);
+
+  .search-result__highlighted,
+  .search-result__teaser {
+    margin: var(--size-spacing-5) 0;
+  }
+
+  .search-result__teaser {
+    color: var(--color-gray-600);
+  }
```
joetower, line 28: This is very minor, but I suggest moving these nested selectors outward.

So it reads more like this: 
 
```
.search-result {
  @include atoms.clickable-component-heading-link;

  padding: var(--size-spacing-5) 0;
  border-bottom: 1px solid var(--color-gray-300);

  @media (min-width: tokens.$break-mobile-max) {
    max-width: var(--size-component-layout-width-content);
  }
}

.search-result__highlighted,
.search-result__teaser {
  margin: var(--size-spacing-5) 0;
}

.search-result__teaser {
  color: var(--color-gray-600);
}
  ```

Comment (joetower) [reply components/02-molecules/search-result/_yds-search-result.scss:28]: Also, I notice that on hover the search result doesn't get a bottom box shadow like the design: https://www.figma.com/proto/G4sJ0BLu4jCJpGzW82JHE8/YaleSites-Components?node-id=6288%3A19237&scaling=min-zoom&page-id=6288%3A19236&hide-ui=1

I'll defer to Ashley on the design review here, but if we do want the bottom-only shadow on hover, there are 2 other instances in the codebase where this kind of hover box-shadow is in place. 

We have used the following (with the TODO comment in place): 
```
    &:hover {
      // @TODO: Temporary styles until we create a token
      // This shadow is hard coded because it has to be adjusted to make it
      // appear as though it's a shadow on only the border-bottom element.
      // There isn't currently anything else that looks like this. If there ever
      // Is, perhaps it should be captured as a token in a new drop-shadow set.
      // @LINK: https://yaleits.atlassian.net/browse/YALB-844.
      box-shadow: 0 8px 6px -6px hsl(0deg 0% 0% / 16%);
    }
```

**Review by joetower — CHANGES_REQUESTED** [review]
@marcb4k I think this is very close. I have a minor CSS comment and I'll tag @ahitson  here for design review.

**Comment by ahitson** [comment]
- Make the entire breadcrumb series (/s too) the same inactive color.
- When it switches to mobile view, keep the full large-screen breadcrumb series and just let the lines stack. 
- The linked page title should be in YaleNew H4 size. The underline is also sticking on hover.
- The title and word example should be in Grey 800.

File: components/02-molecules/search-result/_yds-search-result.scss (hunk 1 above)
Comment (marcb4k) [reply components/02-molecules/search-result/_yds-search-result.scss:28]: OK, @joetower, I fixed both of your comments above and added the @TODO so we know where to replace that once there is a token.

As for Ashley's comments below, I've fixed the last two on her list, but her comments about the breadcrumbs I would like to defer to you on this one. I'm happy to learn best practices on this, but from what Brian taught me, those breadcrumb changes shouldn't live in the search result sass, but instead in the breadcrumb sass - though they are only really used (for now) on the search result.

On line 11 of _yds-breadcrumbs.scss, Brain showed me that we could override breadcrumbs with no bottom margin that is being using on search results, but I'm not sure we want to use that same area for overriding the slash color and the mobile changes. Let me know your thoughts. Thanks!

**Comment by marcb4k** [comment]
@ahitson - What do you mean by "The underline is also sticking on hover."? Thanks!

File: components/02-molecules/search-result/_yds-search-result.scss (hunk 1 above)
Comment (joetower) [reply components/02-molecules/search-result/_yds-search-result.scss]: @marcb4k I would probably add: 
```
.search-result & {
    color:var(--color-basic-brown-gray);
  }
```
to the `03-organisms/menu/breadcrumbs/_yds-breadcrumbs.scss` file inside of the `.breadcrumbs__item {` codeblock, maybe at the bottom? 

```
.breadcrumbs__item {
  @include tokens.body-s;

  position: relative;
  flex-shrink: 0;
  display: flex;
  align-items: center;

    [data-breadcrumbs-overflow='hidden'] & {
      display: none;
    }
  }

  &:not(:last-child) {
    margin-right: var(--size-spacing-5);
    padding-right: var(--size-spacing-5);

    &::after {
      content: '|';
      position: absolute;
      right: 0;
      transform: skew(-15deg) translateX(50%);
    }
  }

  .search-result & {
    color:var(--color-basic-brown-gray);
  }
}
```

This way it is all still contained in the same file. Looks like the `breadcrumb__item`s are using the standard body text color, while the links are `var(--color-basic-brown-gray);` so making the items the same color while in the search-result instance seems like a solid approach to me. 

Am I understanding the feedback correctly, @ahitson ?

ahitson: Yep!

ahitson:
@marcb4k it's either not doing the underline thing anymore or I was crazy. But it does look kind of low right now. Should be cutting through that descender on the g.

components/02-molecules/search-result/_yds-search-result.scss
marcb4k: @ahitson - thanks for the feedback. I've made the changes and you can review here:

https://deploy-preview-176--dev-component-library-twig.netlify.app/?path=/story/molecules-search-result--search-result

As for the underline on hover, take a look at the custom card here, and add a lowercase "g" to the title and it happens there too - the hover line is below the "g".

https://deploy-preview-176--dev-component-library-twig.netlify.app/?path=/story/molecules-cards--custom-card

If we want to standardize on the line going through the "g", let us know. Thanks!

ahitson:
Let's tighten it up a little and change the margin spacing to 3 for .search-result__highlighted (or whatever parent is controlling it).

The underline has been bugging me in general. Ideally, it would be like a normal underline marker that cuts through descenders with visual space around them, more like this 
Screen Shot 2022-12-01 at 9 59 43 AM

If you have any ideas, I can make it into a ticket.

marcb4k:
@ahitson - I think we'll need to move the underline issue to another ticket. One of the problems is that it is an animated underline which limits what we can do. So we have:

1. If we switch away from animated underline, we can get what you want with a standard `text-decoration: underline;`
2. If we want to keep animated underline, we can do it with a text shadow, but the "cut off" is minimal - see screenshots.
3. Also keeping the animation, we can have the line go through the letter that goes below the line. See screenshots.

Animated Strikethrough:
animated-strikethrough
Animated text-shadow g:
animated-text-shadow-g
Animated text shadow j:
animated-text-shadow-j
Non-animated text-decoration g:
non-animated-text-decoration-g
Non-animated text decoration j:
non-animated-text-decoration-j

ahitson:
@marcb4k Thank you so much for the underline info, I can make a ticket with that intel as the starting point.

joetower (approved):
@marcb4k This looks good to me. :clap: Nice work!

joetower:
@ahitson Would you call this design-approved for the time being? With that descender underline ticket in mind?

ahitson:
Just waiting on the spacing change!

marcb4k:
@ahitson - Spacing updated. Review here: https://deploy-preview-176--dev-component-library-twig.netlify.app/?path=/story/molecules-search-result--search-result
Thanks!

ahitson (approved):

PR #177, opened 2022-11-21
YALB-824: Feedback: Small size header border shrink

AnaBZC:
### Description of work
- Fix the border botom in the header in mobile version so that it extends to full width 

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-177--dev-component-library-twig.netlify.app/?path=/story/organisms-site-header--header)

### Functional Review Steps
- [ ]  Go to [header](https://deploy-preview-177--dev-component-library-twig.netlify.app/?path=/story/organisms-site-header--header) and check that in mobile mode the border bottom extends to the edges of the page.

### DIFF
--- a/components/00-tokens/layout/_layout.scss
+++ b/components/00-tokens/layout/_layout.scss
@@ -36,7 +36,9 @@ $layout-widths: map.deep-get(tokens.$tokens, size, component-layout, width);
 
 [data-component-width] {
   &:not([data-embedded-components] [data-component-width]) {
-    padding-inline: var(--size-spacing-site-gutter);
+    @media (min-width: tokens.$break-l) {
+      padding-inline: var(--size-spacing-site-gutter);
+    }
   }
 }
 
--- a/components/00-tokens/layout/_layout.scss
+++ b/components/00-tokens/layout/_layout.scss
@@ -36,9 +36,7 @@ $layout-widths: map.deep-get(tokens.$tokens, size, component-layout, width);
 
 [data-component-width] {
   &:not([data-embedded-components] [data-component-width]) {
-    @media (min-width: tokens.$break-l) {
-      padding-inline: var(--size-spacing-site-gutter);
-    }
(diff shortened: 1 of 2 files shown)


joetower (changes requested):
@AnaBZC This change affect the entire site layout padding. 

Note how the spacing here has been affected: https://deploy-preview-177--dev-component-library-twig.netlify.app/?path=/story/page-examples-news--news-article

yalb-824-no-padding




Compare that to production: 
https://yalesites-org.github.io/component-library-twig/?path=/story/page-examples-events--event-list

We don't want to wrap the padding with a media query in the `_layout.scss` file. Please revert that change. 

Instead, for this ticket, please take a look at the `/components/03-organisms/site-header/_yds-site-header.scss` file and add : 
``` 
margin-inline: calc(var(--size-spacing-site-gutter) *-1);
```
To the `site-header__mobile-header` code block on line `68`. Since we only want to affect the mobile version of the header. We can outdent this element using negative margin on the left/right. 

Which should then look like this on the frontend:
yalb-824

joetower:
@ahitson Do you want to give this a design review? Just want to make sure our solution is in line with your design thinking. 
https://deploy-preview-177--dev-component-library-twig.netlify.app/?path=/story/organisms-site-header--header

ahitson:
Line looks fixed! Not seeing hover transitions anymore, is that due to something not being merged?

joetower:
@ahitson The hovers should be back. When that happens we have to clear the cache and rebuild Netlify instance. Could you also please review Percy here, too. 

https://deploy-preview-177--dev-component-library-twig.netlify.app/?path=/story/organisms-site-header--header

joetower (approved):
@AnaBZC Thank you for making that change. This looks good to me. 

Pending Ashley's design feedback this looks ready to merge.

ahitson (approved):

PR #181, opened 2022-12-09
YALB-840 - Grand Hero: Video background

joetower:
### Description of work
- Adds Background Video component: https://deploy-preview-181--dev-component-library-twig.netlify.app/?path=/story/atoms-video-background--video-background
- Adds video option to the Grand Hero: https://deploy-preview-181--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--grand-hero-banner&args=bgColor:blue-yale;withVideo:true
- Makes sure to respect browser/OS settings to reduce motion in its autoplay behavior

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-181--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--grand-hero-banner&args=withVideo:true)

### Functional Review Steps
- [ ] Verify the play/pause button functions as expected
- [ ] Verify that if you disable autoplay in your browser and refresh the page, the background video does not autoplay
- [ ] Verify that if you set reduce motion in your operating system settings and refresh the page, the background video does not autoplay
- [ ] Verify that the background video and its button element are responsive

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/file/G4sJ0BLu4jCJpGzW82JHE8/YaleSites-Components?node-id=6517%3A22048&t=x6nSimcEUIykNVFq-0)

### Accessibility Review


ahitson:
- The play/pause button needs a hover state. Please make the background of the button fade into a solid version of the color on hover/focus.
- Let's use the play icon from the "solid" set. Not a fan of the outlined version in the "regular" set. 
- When the play/pause button shifts to the corner on smaller screens, please have it also reduce the height/width to 1.4 rem.
- Will the motion come once we get it in a page example?

joetower:
@ahitson Awesome! You're so quick! 

I'll make these changes: 
>The play/pause button needs a hover state. Please make the background of the button fade into a solid version of the color on hover/focus.
Let's use the play icon from the "solid" set. Not a fan of the outlined version in the "regular" set.
When the play/pause button shifts to the corner on smaller screens, please have it also reduce the height/width to 1.4 rem.

For this one: 
> Will the motion come once we get it in a page example?
Are you referring to the parallax effect? If so, we have that as a separate ticket: https://yaleits.atlassian.net/browse/YALB-839

ahitson:
- Close! Only the play icon needed to change sets. Can you change the pause back to regular? 
- Also looks like the solid pause miiiiight be visually large compared to the thinner pause button, so we may need to bring the size down  -.3 for just the pause icon...but not sure yet.

ahitson:
Yep, flipping between pause and play is a little too dramatic. Is there a way to just bring the play icon down a bit while keeping the background size the same? It does look like it's bigger than the pause icon.

components/01-atoms/video-background/yds-video-background.twig
```diff
@@ -0,0 +1,39 @@
...
+ #   the "|raw" filter is applied so that the iframe is rendered instead
+ #   of simply passing the iframe as plain text.
+ #
+ # Available blocks:
+ # - video_background__content - used to replace the content of the video with something other than the typical iframe
+ #   for example: to insert an html5 video component
+#}
+
+{% set video_background__base_class = video_background__base_class|default('video-background') %}
+{% set video_background__button__background_color = video_background__button__background_color|default('gray-800') %}
```
callinmullaney, line 13: Add this variable to the "Available Variables" comment above with the available options. I believe the "Callout" component has an example of this.

joetower:
@ahitson It definitely carries more visual weight because it is a more dense and solid graphic in comparison to the rather sparse pause icon.

How does this balance feel? I added some padding to the sides of the play icon SVG. 

https://deploy-preview-181--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--grand-hero-banner&args=withVideo:true

components/02-molecules/banner/grand-hero/_yds-grand-hero.scss
```diff
@@ -30,13 +30,15 @@ $break-grand-hero-banner: tokens.$break-m;
   }
 }
 
-.grand-hero-banner__image {
+.grand-hero-banner__image,
```
callinmullaney, line 33: I'm surprised adding a new class here with a comma didnt cause linting issues with specificity in the nested elements. At any rate, the combination of these class with video,img will result in styles that never apply. Example: `.grand-hero-banner__image video {}` - I would probably change these wrapping classes to `.grand-hero-banner__media` with a video or img modifier if you need to separate these styles. Not a big deal but thought I would mention this other approach to simplify these styles.

components/02-molecules/banner/grand-hero/yds-grand-hero.twig
```diff
@@ -17,6 +17,7 @@
 {% set grand_hero__content__background = grand_hero__content__background|default('gray-800') %}
 {% set grand_hero__overlay_variation = grand_hero__overlay_variation|default('contained') %}
 {% set grand_hero__size = grand_hero__size|default('full') %}
+{% set video_background__button__background_color = grand_hero__content__background|default('gray-800') %}
```
callinmullaney, line 20: I think this variable should be set at the include level and not at the top here. Otherwise, you would need to add `video_background__button__background_color` as an available variable at the top of this file and that seems inconsistent with the other components.

callinmullaney (changes requested):
@joetower Nice work on this one! Functionally everything looks great. I had a couple of code formatting comments that shouldn't effect rendering but are more inline with other organization that was established in other components. Feel free to push back on these if you disagree. 😄

ahitson:
Much better!

ahitson (approved):

components/02-molecules/banner/grand-hero/_yds-grand-hero.scss
joetower: @callinmullaney I just went ahead and separated them. :+1:

joetower:
@callinmullaney Thank you so much for the great feedback! Could you please give this another look, all of your feedback should be addressed. :+1:

callinmullaney (approved):
Looks like my feedback was addressed. PR looks good to me!

callinmullaney:
@joetower I just approved but had a thought I wanted to pass on. Shouldn't be a blocker for merging if you choose not to follow through. With the Atoms section in storybook I think we should group these two Video type components into a single folder. Then have Background and Embed as the options below. I don't think you should combine the actual twig files or anything like that but group them together in storybook. Similar to how Lists are grouped but you have the two options below.

joetower:
@callinmullaney Great idea! I added the videos organization and will merge this. Thank you!

PR #185, opened 2023-01-10
YALB-913: Bug: Search results page

ShaneMcPartlinYale:
### Description of work
- Added padding underneath search page form
- Increases height of text input within search page form
- Before: 
image
- After:
image
- Companion PR in atomic: https://github.com/yalesites-org/atomic/pull/79

##Design review
- Figma:
image
- This branch:
image

### DIFF
--- a/components/01-atoms/forms/textfields/_yds-textfields.scss
+++ b/components/01-atoms/forms/textfields/_yds-textfields.scss
@@ -20,10 +20,6 @@ $form-item-icon-size: 1.7rem;
   }
 }
 
-.search-form--page.form--inline {
-  margin-bottom: var(--size-spacing-8);
-}
-
 .form-item__label {
   display: block;
   margin-bottom: var(--size-spacing-1);
--- a/components/01-atoms/forms/textfields/_yds-textfields.scss
+++ b/components/01-atoms/forms/textfields/_yds-textfields.scss
@@ -14,6 +14,14 @@ $form-item-icon-size: 1.7rem;
   margin-bottom: var(--size-spacing-7);
   max-width: $form-item-max-width;
   clear: both;
+
+  .form--inline & {
+    margin-bottom: unset;
+  }
+}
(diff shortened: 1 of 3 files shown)


components/01-atoms/forms/textfields/_yds-textfields.scss
```diff
@@ -14,6 +14,14 @@ $form-item-icon-size: 1.7rem;
...
   clear: both;
+
+  .form--inline & {
+    margin-bottom: unset;
+  }
+}
+
+.search-form--page.form--inline {
+  margin-bottom: var(--size-spacing-8);
 }
```
joetower, line 25: @ShaneMcPartlinYale This is super minor, but I would remove this code block
```
.search-form--page.form--inline {
  margin-bottom: var(--size-spacing-8);
}
```

And apply the margin-bottom in the search-result file here: `/component-library-twig/components/02-molecules/search-result/_yds-search-result.scss`(

And simply add `margin-bottom: var(--size-spacing-8);` on line 13, after the `gap` property, like this: 
```
.search-form--page {
  display: flex;
  gap: var(--size-spacing-5);
  margin-bottom: var(--size-spacing-8);
}
```

joetower (changes requested):
@ShaneMcPartlinYale Very minor change suggestion here. Otherwise it's looking good. :+1: 

For Ashley's feedback on adjusting the amount of space between the `page-title` element and the `search-results--summary` element, I'd like to address that feedback in a future ticket. I want to give some thought to how we might best implement spacing deviations in instances like this.

components/01-atoms/forms/textfields/_yds-textfields.scss
ShaneMcPartlinYale: I have made these changes in the latest commit. Thanks for pointing it out!

joetower (approved):

PR #186, opened 2023-01-11
YALB-920 - Forms: Theming next steps

joetower:
### Description of work
- Updated component wrapper to include a title
- Adds a fix for `form-item`s inside the `utility-nav` (the search input) to remove bottom-margin

### Functional Review Steps
- [ ] Test in the Atomic repo here: https://github.com/yalesites-org/atomic/pull/80


components/03-organisms/component-wrapper/yds-component-wrapper.twig
```diff
@@ -1,6 +1,19 @@
-<div data-component-width={{ component_width|default('max') }}>
+{% set component__base_class = 'component-wrapper' %}
+
+<div {{ bem(component__base_class, component__modifiers, component__blockname, component__extra_classes) }} data-component-width={{ component_width|default('max') }}>
```
callinmullaney, line 3: Whenever you need to tackle more attributes besides classes I'd suggest using `add_attributes` instead. This will make the component much more extendible by other components that use it. Something like ...
```
{% set component_wrapper__attributes = component_wrapper__attributes|default([]) %}

{% set component_wrapper__attributes = component_wrapper__attributes|merge({
  'class': bem(component__base_class, component__modifiers, component__blockname, component__extra_classes),
  'data-component-width': component_width|default('max'),
}) %}

<div {{ add_attributes(component_wrapper__attributes) }}>
```

components/01-atoms/forms/_form-element.twig
```diff
@@ -0,0 +1,88 @@
+{#
```
callinmullaney, line 1: This file in consistent with other projects we use but is very drupal-y. Not saying something needs to change here to merge this but it would be nice to figure out a different solution or something that is more inline with the other files. Just my two-cents :stuck_out_tongue:

callinmullaney (changes requested):
@joetower See code comment about the component_wrapper. Also, that file is using variables like `component_label`. In order to keep the same convention established in other files I think these should be changed to `component_wrapper__label`.

components/03-organisms/component-wrapper/yds-component-wrapper.twig
joetower: @callinmullaney Ah, yeah, good call! Thanks!

components/01-atoms/forms/_form-element.twig
joetower: @callinmullaney Do you have ideas as to how to make it less Drupal-y? Being that we need form validation from Drupal, I'm not sure how we'd do forms differently.

nJim: Does it make sense to move any of this to a template file in the atomic-theme? I know that our wiring-files are usually pretty superficial and most of the work is done in other places. But forms sit in that special place where the markup implies a lot of functionality.

Of if you want me to do any custom field logic in PHP, I can alter elements/attributes in the ys_theme module?

joetower: @callinmullaney @nJim I suppose we could remove the `_form-element.twig` and `_form-label.twig` files in the component library and instead of `{% include`-ing them in the Drupal template, simply use the markup instead. What are your thoughts? That way at least it's not in the component library. 

We really just wanted a way to show what forms should look like in Storybook.

nJim (approved):
Approved.

PR #187, opened 2023-01-11
YALB-908 - Bug: Text link style bug for Callout

joetower:
### Description of work
- fix: `a` element coloring in callouts/cards
- The scope of this ticket was for callouts, but I figured we might as well do the same for custom cards, just in case :+1: 

### Testing Link(s)
- [ ] This PR includes changes to the callout and card components, but we don't have an example in the component library for `a` elements used inside of `callout` or `card` content. 

### Functional Review Steps
- [ ] To test these changes, you can either test by checking out this branch and using Drupal to add/edit a page node (or test in Storybook, details below), and add a `callout` and `card` component to the content. If you do that, you should see this: 

yalb-908

- [ ] You could also test these changes by navigating to: https://deploy-preview-187--dev-component-library-twig.netlify.app/?path=/story/molecules-callout--callout
- [ ] And, using the controls options in Storybook, pasting a link in the `text` content, like this: 
`<p>Designed for those who intend to pursue graduate and those who wish to <a href="https://google.com">immediately enter a career</a> in which broad scientific training is beneficial.</p>`
- [ ] Then you could do the same for the `custom-card`: https://deploy-preview-187--dev-component-library-twig.netlify.app/?path=/story/molecules-cards--custom-card 
yalb-908-hover-storybook

### DIFF
--- a/components/02-molecules/callout/_yds-callout.scss
+++ b/components/02-molecules/callout/_yds-callout.scss
@@ -84,11 +84,7 @@ $callout-themes: map.deep-get(sass-tokens.$tokens, callout-themes);
   @include tokens.body-default-condensed;
 
   a {
-    color: var(--color-text);
-  }
-
-  a:hover {
-    text-decoration: none;
+    @include atoms.link;
   }
 
   p:last-child {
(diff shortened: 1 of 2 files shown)


nJim (approved):
Approving this work from my point of view. The change seems pretty standard. I like the specificity of adding styles that are scoped to the component as I think it makes things more future proof. But feel free to wait for Callin's feedback if you have bigger questions.

PR #48, opened 2022-03-22
feat(YALB-69): icon support (including webpack setup)

ModulesUnraveled:
### Description of work
- Creates the icon component
- Configures Webpack to automatically build our icon sprite based on svgs in the `/images/icons` directory

### Testing Link(s)
- [ ] Navigate to the [Icons Story](https://deploy-preview-XX--dev-component-library-twig.netlify.app/?path=/story/atoms-images--icons)

### Functional Review Steps (to be done locally)
- [ ] `npm install`
- [ ] `npm run develop`
- [ ] Verify the four initial icons are available
- [ ] Download any FontAwesome (or other) icon ([like this one](https://fontawesome.com/icons/infinity?s=solid))
  - [ ] Drop the icon into the top level `/images/icons` directory (next to the existing four icons)
  - [ ] Verify the icon is automatically added to the Icons story in Storybook

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/file/OVfmCW5s1gI7xnrM8ibX2y/UI-Kit-Alpha-%5BYaleSites%5D)

### Accessibility Review
- [ ] Verify the component meets Accessibility requirements

## Notes:
- I'm currently using the "solid" version of icons, but at least a couple will need to be replaced by the "regular" version once I have access to the pro account.

PR #195, opened 2023-01-19
fix: new link__type to denote external and new-target link types

joetower:
### Description of work
- Adds font-awesome
- Uses font-awesome for new text-link styles: https://deploy-preview-195--dev-component-library-twig.netlify.app/?path=/story/atoms-controls--text-link
  - external
  - new target
  - download
  - with chevron

### Testing Link(s)
- [ ] Navigate to the [Component Story]( https://deploy-preview-195--dev-component-library-twig.netlify.app/?path=/story/atoms-controls--text-link)

### Functional Review Steps
- [ ] Review the link types:  https://deploy-preview-195--dev-component-library-twig.netlify.app/?path=/story/atoms-controls--text-link
- [ ] And new fa-icon component: https://deploy-preview-195--dev-component-library-twig.netlify.app/?path=/story/atoms-images--fa-icons 

### Design Review
- [ ] Verify the designs match the [Notion Designs in Documentation](https://www.notion.so/fourkitchens/Links-afa9aa31f9544248b7bbe529f9c02dc8)

### Accessibility Review
- [ ] @ntay90 I am curious as to whether our `svg` approach to icons is more accessible or at the same level as the Font Awesome font-based `<i>` approach.  Could you please review? Additional details below, but let me know if you have any questions. :+1: 
  - [ ] Compare the accessibility of our SVG-based icons: https://deploy-preview-195--dev-component-library-twig.netlify.app/?path=/story/atoms-images--icons
  - [ ] SVG in real component use case example here: https://deploy-preview-195--dev-component-library-twig.netlify.app/?path=/story/molecules-alert--alert-examples
  - [ ] With the new font-based approach here: https://deploy-preview-195--dev-component-library-twig.netlify.app/?path=/story/atoms-images--fa-icons
  - [ ] font-based in real world use case example here: https://deploy-preview-195--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--grand-hero-banner (This is a link >)
  - [ ] I would like to assess whether we can remove the SVG icons, refactoring each in-use example to use the font-based approach, if possible. Otherwise, we could keep both but use the font-based approach sparingly. We want to use the font-based approach for some WYSIWYG editor link icons, as they would be easier to implement than the SVG approach.


ahitson:
Let's change the external icon to "solid" so it matches the new tab icon.

kara-franco:
heya @joetower 
jumping in here because @ntay90 is out sick. love the question about method for icons. we chatted about this in PR #48. We also discussed in Teams (i'll continue searching for the convo, it was archived and i'm having trouble finding it). no matter the method, we will need to make the icons for the chevron decorative (use of aria-hidden=true) AND make the icons for the special links make sense for screen reader users. 

from our research, the SVG method is the easiest to add semantic meaning to icons and when needed, make them decorative. if we go the icon `<i>` font-based approach, we will need to add <span> with a sr-only class to provide information to screen reader users, further reading: [semantic icons](https://fontawesome.com/v5/docs/web/other-topics/accessibility#web-fonts-with-css-semantic-icons). 

currently, the font-based approach for the links are not accessible. for a screen reader user, they do not hear the extra meaning that the icons are giving (question mark icons in web rotor in screenshot). for example, they should hear "This is an external link, external link", but currently they only hear the link text. 

Screen Shot 2023-02-06 at 11 24 30 AM

joetower:
@kara-franco Thank you! So, maybe there is a solid use case to keep our SVG workflow and use these FontAwesome instances more sparingly. 

I pushed up a couple changes. One to pass in the decorative = true variable, so that we get the aria hidden attributes added to the `<i>` element along with the `role`. Also, I added a screen reader-specific title for the `<i>` element. 

If you could please take another look, you should see the following, as shown in my screenshot below . I'm curious, however, if the `title` attribute and the `aria-describedby` are redundant - are both necessary or is one more favorable than the other? What are your thoughts?  Thank you! 

https://deploy-preview-195--dev-component-library-twig.netlify.app/?path=/story/atoms-controls--text-link

Screenshot-20230206105146-1536x1328

kara-franco:
@joetower i just listened to the links with VoiceOver and still cannot hear the extra info that the icons are giving. Image below is what is read to screen reader users.

Screen Shot 2023-02-06 at 1 33 24 PM

We should hear: 
"This is an "external" link, **(link is external)**"
"This is a "new target" link, **(opens in a new tab)**" 
"This is a "download" link, **(file download)**" 
"This is a link with chevron"

Currently the external, new target and download links do not have the extra info read out. We should either use the svg method or add in a `<span>` that is screen reader only. Example: 
`<a my cool link> 
 <i aria-hidden="true" class="fas fa-car"></i>
  <span class="sr-only">(opens in a new tab)</span>
</a>`

Stanford uses the SVG method: 
Screen Shot 2023-02-06 at 1 29 37 PM

Last, we want to avoid the title attribute because it adds in a tooltip, which is undesirable visually and not needed accessibility-wise. 
Screen Shot 2023-02-06 at 1 12 27 PM

joetower:
@kara-franco Thank you so much! 

Our goal is to use the `<i>` Font Awesome font for any icon which gets rendered with WYSIWYG text content. For everywhere else on the site, we can use the `svg` implementation, but the `svg` implementation would be more difficult than if we could use the icon font Font Awesome provides.  

I have changed the previous fix I pushed, and now you should see the following, which I hope gets read as it should by screen readers. Could you please give this another look. Thank you!
Screenshot-20230206133800-1407x1333

kara-franco:
@joetower Awesome the three non-decorative links sound great. The chevron link is decorative, screen reader users don't need to hear the extra info that it's decorative. So we can remove the sr-only span and just stick an aria-hidden on it. 

` 
 <i class="fa-icon fa-solid fa-angle-right" aria-hidden="true"
    </i>`

joetower:
@kara-franco Thank you for the back-and-forth on this ticket, I appreciate it!

I pushed a fix to remove the sr text from the link with chevron. Please take another listen. Thank you!

kara-franco:
@joetower looks perfect! have a lovely vacation!

joetower:
@kara-franco Thank you so much :smile:

PR #198, opened 2023-01-31
YALB-980: Adjust component-width

joetower:
### Description of work
- fix(yalb-980): adjust component-width for some components and add alignment for text
- assigning the `width` data attribute in several components to a variable so that we can pass different values in Drupal rather than the value being hardcoded at the component level

### Functional Review Steps
- [ ] Test here https://github.com/yalesites-org/yalesites-project/pull/208


nJim (approved):
Approved. Works well on my local. We will skip the drupal 'max'/left-align part for now. But this would should be able to move forward without it.

PR #249, opened 2023-06-07
YALB-1300: Bug: Error when creating a new event

codechefmarc:
### Description of work
- Do not show date and time if they are not set
- Fixes showing current date if date and time are not set

### Testing Link(s)

### Functional Review Steps
- [ ] Full testing at [PR-302](https://github.com/yalesites-org/yalesites-project/pull/302)

### Design Review

### Accessibility Review


dblanken-yale (approved):
This looks pretty good.  The only thing bugging me with this change in Storybook is seeing "the event has passed" when no dates have been entered or are showing.  Would this work as a replacement of line 32/33?

```twig
{# Check if event has an end date #}
{% if event_meta__date_end %}
  {# Check if the date has passed the end date #}
  {% set event_has_passed = event_meta__date_end < "now"|date("Y-m-d") %}
{# Check if the event only has a start date #}
{% elseif event_meta__date_start %}
  {# Check if the date has passed the start date #}
  {% set event_has_passed = event_meta__date_start < "now"|date("Y-m-d") %}
{# If there is no start or end date, the event has not passed #}
{% else %}
  {% set event_has_passed = false %}
{% endif %}
```

That way, if the end date exists, it does like it did before, but if it doesn't and only a start date exists, it evaluates against that date, and if no dates exist, it does not evaluate the time comparison?  Does that sound reasonable or can you think of a use case in the project that it could pose a problem for?

As far as what you're trying to fix and the scope of the work, all works as expected otherwise.  I just noticed that when I was testing and didn't know if we could get that in, too, or if that should be a separate ticket.

Approving since your changes do what you set out to do, but if you agree that change would be helpful, go for it!

PR #259, opened 2023-06-28
YALB-1418: Bug: All events created are showing as past events

dblanken-yale:
### Description of work
- Change date format of `now` to use unix timestamps over human readable

### Testing Link(s)
- [ ] Navigate to the [Event Meta Component Story](http://localhost:6006/?path=/story/molecules-meta--event)
- [ ] Navigate to the [Event Page Component Story](http://localhost:6006/?path=/story/page-examples-events--event-page)
- [ ] Use a local YaleSites instance to test that it works with Drupal's data as well.

### Functional Review Steps
- [ ] In Storybook links
  - [ ] Change the dates to ensure that it only says past event when it is a past event
- [ ] In local YaleSites instance
  - [ ] Create a new event, making the date/time occur in the future, and save
  - [ ] Verify that it does not say it is a past event
  - [ ] Create or modify an event, making the date/time occur in the past, and save
  - [ ] Verify that it does say it is a past event

### DIFF
--- a/components/02-molecules/meta/event-meta/yds-event-meta.twig
+++ b/components/02-molecules/meta/event-meta/yds-event-meta.twig
@@ -30,7 +30,7 @@
 {% set event_meta__base_class = 'event-meta' %}
 
 {# Check if event has passed #}
-{% set event_has_passed = event_meta__date_end < "now"|date("Y-m-d") %}
+{% set event_has_passed = event_meta__date_end < "now"|date("U") %}
 
 {# Set the event status based on event_has_passed variable #}
 {% set event_status = event_has_passed ? 'passed' : 'current' %}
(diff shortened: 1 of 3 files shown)


joetower (approved):
@dblanken-yale This looks great to me, nice work :+1: 

I did notice, however, that the `atomic` side of the "Past Event" display was never completed (part of a previous ticket which may not have included an Atomic branch). 

When I tested your PR, here, locally, in Drupal, I was seeing an accessibility error because the Atomic Template for the Full Event Node was still pulling in the page title content with the event meta inside of it. Because we changed the Event Meta Component to include the page_title, we need the Event Meta component to get pulled in, instead, with the `event_title__heading` mapping to the `label` so that the page title gets included inside the Event Meta component. Please let me know if you have any questions. I searched the Jira backlog for an Atomic side of the Past Event work or a bug related to what I just added in Atomic and didn't find anything.

So, I added a branch with the same branch name as this one you've opened and asked for your review here: https://github.com/yalesites-org/atomic/pull/141. Apologies if you were working on this already. I just figured I'd go ahead and add it in. :+1: 

Thank you!

dblanken-yale:
@joetower No that was great.  Great catch!  I'll take a look at it right now.  Thanks for looking, finding that extra, and letting me know.  It will be good to have both of these in before Friday for the testers!

PR #261, opened 2023-06-30
YALB-1326): Bug: Accessibility issues with figure captions

codechefmarc:
### Description of work
- Removes the `aria-label` from `figure` tags as the caption is already in the DOM and screens readers were reading the HTML tags in the `aria-label`

### Testing Link(s)

### Functional Review Steps

### Design Review

### Accessibility Review


myvaughn (approved):
Tested figure without aria-label in macOS VoiceOver/Safari and NVDA/Chrome, and they announced the caption when entering the figure as expected. I didn't test JAWS but expect it to work fine. It does not work with Narrator, and does not work for iOS VoiceOver. But I think using this standard HTML is fine, and hopefully browser support will eventually catch up.

PR #274, opened 2023-07-14
YALB-1394: Bug: Accessibility (FE)- Empty heading being inserted when user does not fill in heading field

dblanken-yale:
## [YALB-1394: Bug: Accessibility (FE)- Empty heading being inserted when user does not fill in heading field](https://yaleits.atlassian.net/browse/YALB-1394)

Also linked to [Atomic PR 152](https://github.com/yalesites-org/atomic/pull/152) in order to pass the correct heading values in certain components.

Thanks to @joetower for showing me the beauty of creating a temp multidev that can use these PRs!
[Multidev to test with](https://pr-357-yalesites-platform.pantheonsite.io)

### Description of work
- Change presence `if` statements from checking for the existence of the variable to checking that the variable is not empty in the following places:
  - yds-accordion.twig
  - yds-grand-hero.twig
  - yds-video.twig
  - yds-card-collection.twig
  - yds-custom-card-collection.twig
  - _yds-media-grid-modal-item--content.twig
  - yds-component-wrapper.twig
  - yds-media-grid.twig

### Testing Link(s)
- [ ] Navigate to each of the component stories
  - [ ] [Accordion](https://deploy-preview-274--dev-component-library-twig.netlify.app/?path=/story/molecules-accordion--accordion)
  - [ ] [Grand Hero](https://deploy-preview-274--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--grand-hero-banner)
  - [ ] [Video](https://deploy-preview-274--dev-component-library-twig.netlify.app/?path=/story/molecules-video--video)
  - [ ] [Custom Card](https://deploy-preview-274--dev-component-library-twig.netlify.app/?path=/story/molecules-cards--custom-card)
  - [ ] [Custom Card Collection](https://deploy-preview-274--dev-component-library-twig.netlify.app/?path=/story/organisms-card-collection-custom-card-collection--custom-card-collection)
  - [ ] [Image Grid](https://deploy-preview-274--dev-component-library-twig.netlify.app/?path=/story/organisms-galleries--image-grid)
  - [ ] [Interactive Grid](https://deploy-preview-274--dev-component-library-twig.netlify.app/?path=/story/organisms-galleries--interactive-grid)
  - [ ] [Standard Pages with Banner](https://deploy-preview-274--dev-component-library-twig.netlify.app/?path=/story/page-examples-standard-pages--with-banner)

### Functional Review Steps
- [ ] Verify that when setting the header to nothing that the headings do not render in the DOM (if you're able to set the header)

### Design Review

### Accessibility Review
- [ ] Verify the components still meet Accessibility requirements without the heading


codechefmarc (commented):
I'll defer to @joetower for most of this PR as this is more in his world. I tested all of the components and most worked great, with a few exceptions:

1) Accordion works (to change accordion items from h3 to h2 when no accordion heading is set), but I'm getting `object(2) { [accordion__item__heading]=> string(15) "Art Collections" [accordion__item__content]=> string(817) "` strings printing inside the accordion items.
2) Custom card - I'm still getting an empty h2 on the card title when the card title isn't set. This "can't happen" (hah) because we require the card heading in Drupal, but as a fail-safe, probably need to fix this one. Issue is, I think, that `a` tag needs to be there as I think its the only one for the whole card? Not sure.
3) For the "Post Article" test (https://deploy-preview-274--dev-component-library-twig.netlify.app/?path=/story/page-examples-post--post-article) I'm not sure what I'm supposed to test here? I removed the page title and an empty h1 still exists, but again that'll "never happen" in Drupal. Also on that page were some custom cards but we already tested that one above.

myvaughn:
@dblanken-yale I tested on my local instance (Drupal, not component library). Notes/failure:

- "Grand Hero" - doesn't allow me to exclude the heading (so could not test)
- "Video" - still results in an empty heading if empty title
- "Image grid" - still results in an empty heading if empty title
- "Interactive grid" (gallery) - has lots of empty headings. For the component and for each image in the modal.
- "Post article" - if this is the heading (title) for a post content type, it's required, so I can't test
- "Action banner" - heading required, can't test

dblanken-yale:
@codechefmarc Thanks so much for looking at it!  And thanks for reminding me to push my removal of the dump command.  hahahahaha

1. After actually committing that removal the dump information should no longer show.
2. I got the same thing when I removed the text and then refreshed the page; however, when I add text back and then remove it and look, it no longer shows.  I'm going to have to look into that a little more--that makes me think it's a storybook issue that we may want to address for consistency?
3. I'll have to check that one as well--I basically did a grep for what used these files to come up with that list, so I'll dive further to how it's actually using that one.

@myvaughn Thanks for looking as well.  I'll see what I can do to add that to storybook so we can test it.  Even if the drupal side has it as required, it makes sense that the component should be flexible and allow us to see these changes since it's a parameter you can change.  I'll dive into the atomic side for the ones that still showed it.  I suspect a similar thing is happening that was happening with the accordion where the render is being overwritten.

dblanken-yale:
@codechefmarc My bad.  The post one was due to it using the card collection at the bottom, but we have no way of modifying that.  I think by looking at the card collection it should be good enough.  That's what I get for grepping.  :D

codechefmarc (approved):
Tested in Storybook:
1. The Accordion debug is gone.
2. The custom card still has an H2 when there is no title, but, like we pointed out before, this isn't possible to get to in Drupal and it also provides the `a` tag for the link, so if that is required, I would think that the heading shouldn't be able to be deleted? Or perhaps chat with Joe to see how to refactor that piece if there is no heading, but we still need the link for the card?

Tested in Drupal:
1. All components (that made sense to test) all worked great when removing optional headers.

dblanken-yale:
@codechefmarc True on the custom card.  Looking at the javascript functionality for it, we're targeting the header to do some stuff.  I agree that it should be considered a required field, so I won't modify that.  I think this is ready then.

PR #297, opened 2023-09-19
fix(YALB-email-copy-change-to-button): create new component

joetower:
### Description of work

> HIGH PRIORITY! Copy button for email link in Profile is not receiving focus and is not tabbable- (should be a button and not a link) @joetower

- Adds a new separate component for copy text link so that we can change the element from an `a` to a `button`. Also, it helps make the text-link component much more lean and makes the new component more purposeful. 

#### Includes: 
- https://github.com/yalesites-org/atomic/pull/176
- https://github.com/yalesites-org/yalesites-project/pull/433

### Functional testing steps:
- [ ] Test here: https://github.com/yalesites-org/yalesites-project/pull/433
- [ ] New component is here: https://deploy-preview-297--dev-component-library-twig.netlify.app/?path=/story/atoms-controls--text-copy-button-examples

### DIFF
--- a/components/01-atoms/atoms.scss
+++ b/components/01-atoms/atoms.scss
@@ -1,6 +1,7 @@
 @forward './controls/button/yds-button' as button-*;
 @forward './controls/cta/yds-cta';
 @forward './controls/text-link/yds-text-link';
+@forward './controls/text-copy-button/yds-text-copy-button';
 @forward './date-time/yds-date-time';
 @forward './divider/yds-divider';
 @forward './forms/yds-form' as forms-*;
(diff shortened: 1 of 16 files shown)


dblanken-yale (approved):
Very nice component.  Works great!

I noticed some mega-footer/social-links stuff in this.  Was that on purpose?  Just checking.

joetower:
@dblanken-yale So, I changed the base of this PR to `develop` and then back to `YALB-sept-release-fixes`. That was weird, though. It's as if some git history from `develop` wasn't removing itself? 

It should look less weird now?

nJim (approved):
Approved.

PR #299, opened 2023-09-22
YALB-1527 - Bug: Update link styling "Skip to Main Content" | YALB-1560 - Blank space in toolbar on admin theme

joetower:
### Description of work
- Adds a component for `link skip`: 
- Wires new link skip to `html.html` template.
- Fixes chosen input search colors on `admin/content` tags

### Testing Link(s)
- [ ] Navigate to the [Link Skip Component Story](https://deploy-preview-299--dev-component-library-twig.netlify.app/?path=/story/molecules-link-skip--link-skip)

### Functional Review Steps
- [ ] Test here: https://github.com/yalesites-org/yalesites-project/pull/438

### DIFF
--- a/components/01-atoms/controls/text-link/_yds-text-link.scss
+++ b/components/01-atoms/controls/text-link/_yds-text-link.scss
@@ -95,8 +95,6 @@ $global-link-themes: map.deep-get(tokens.$tokens, 'global-themes');
 // and directly referenced at the bottom of this file in:
 // &[data-link-style='no-underline-animation']
 @mixin plain-link {
-  @include tokens.animate;
-
   // Global themes: set color slots for each theme
   // This establishes `--color-slot-` variables name-spaced to the selector
   // in which it is used. We can map component-level variables to global-level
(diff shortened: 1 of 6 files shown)


joetower:
@dblanken-yale 🤔 That could be tough to add into the component library since it would have to be added it to the `components/04-page-layouts/_yds-base.twig` file, which would render it for any page that includes it (which would include duplicating it in the Drupal templates, unless we remove the skip-link from the html.html template). 

The actual display of the skip link is provided by Drupal in `/core/themes/stable9/css/system/components/hidden.module.css?s1pcoq` using the CSS of `.visually-hidden.focusable:active, .visually-hidden.focusable:focus`. The skip link is hidden by default. 

If we wanted it to work in the component-library we would want to replace this functionality from within our component-library. 

I'm not sure this is what we want to do. 🤔 What are your thoughts?

dblanken-yale:
I think you're right; I mean the component library is displaying the component correctly, and it's trivial to see what it's targeting, so I think it'll be ok.  And we can see down the line if anyone gets mixed up and if so, handle it then.  :D

codechefmarc (approved):
Approved as well!

PR #310, opened 2023-11-29
YALB-1148: Quotes: Add standalone treatment to storybook

joetower:
### Description of work
- Adds a new component for standalone quote, borrowing styles from the pull quote since the design was based on the same quote styles.
- The styles from pull quote were separated into `styles` and `alignment`. Moves the styles are now more single purpose - `bar` and `quote` instead of `bar-left` `bar-right`. `image` is only set as a style if Quote image is set to `with-image` which we can set in Drupal based on the quote image field being populated or not. 
- Alignment is a separate option
- Note: the pull quote did not have a `quote-right` option, so I made the quote right resemble the bar-right option, right-aligned text with the same spacing rules. 
- Moves Pull Quote into the same component Storybook parent directory - so both appear under the "Quotes" folder

### Testing Link(s)
- [ ] Navigate to the [Standalone Quote Component Story](https://deploy-preview-310--dev-component-library-twig.netlify.app/?path=/story/molecules-quotes-standalone-quote--standalone-quote)

### Functional Review Steps
- [ ] Navigate to the new component: https://deploy-preview-310--dev-component-library-twig.netlify.app/?path=/story/molecules-quotes-standalone-quote--standalone-quote
- [ ] Test the control options - theme, alignment, style, with image

https://github.com/yalesites-org/component-library-twig/assets/366413/52156cda-e1ac-430e-b368-cbc86e39f168

- [ ] Take a look at a Standalone Quote on the global color theme example page: https://deploy-preview-310--dev-component-library-twig.netlify.app/?path=/story/tokens-colors--global-theme-color-pairings

https://github.com/yalesites-org/component-library-twig/assets/366413/f0d7939e-3429-44e5-a2ae-19d323db6857

- [ ] View a Standalone Quote on an example page: https://deploy-preview-310--dev-component-library-twig.netlify.app/?path=/story/page-examples-standard-pages--with-banner

### Design Review
- [ ] Verify the designs match the [Spotlight Quote Figma Designs](https://www.figma.com/proto/G4sJ0BLu4jCJpGzW82JHE8/YaleSites-Components?type=design&node-id=7956-25378&t=d9aiDmjddXzx83se-0&scaling=min-zoom&page-id=5218%3A22754)

### Accessibility Review


kara-franco (approved):
Looks awesome, I testing in all color combos and it passes.

dblanken-yale (approved):
@joetower Making sure I am understanding correctly; when you say "image is only set as a style if Quote image is set to with-image which we can set in Drupal based on the quote image field being populated or not." you mean that when "with-image" is used, style is force set to "image", and the quote/bar settings are discarded in that case?  If so, it threw me off to not see that as a style but I get it.

Looks great; just had that one question.

joetower:
@dblanken-yale Yes, that is exactly it. I thought it made the most sense, architecturally, but open to other suggestions if you see a better path. Thank you!

codechefmarc (approved):
Approved here as well - great work! In Drupal, we may need to make sure we have some help text to let folks know that if they have an image, that the "style" (quote vs bar) won't work or we may be able to do some auto-hiding of the style field if "with image" is selected?

PR #313, opened 2023-12-06
YALB-1657 - Update the expand/toggle for image gallery text | YALB-1658 - Update search box styles

joetower:
### Description of work
YALB-1657
- Adds new icons for circle-plus and circle-minus
- Changes the interactive gallery expand/collapse icon to circle-plus/circle-minus
- Move the icon to the left of the caption instead of the right
- Add background color to the caption (if it has a toggle) 

YALB-1658 
- Add a border to the header utility nav search form item
- Move the search icon to the right of the input

### Testing Link(s)
- [ ] Navigate to the [Interactive gallery component](https://deploy-preview-313--dev-component-library-twig.netlify.app/?path=/story/organisms-galleries--interactive-grid)
- [ ] Navigate to the [Utility nav component](https://deploy-preview-313--dev-component-library-twig.netlify.app/?path=/story/organisms-menu-utility-nav--utility-nav)

#### Testing: [YALB-1657 - Update the expand/toggle for image gallery text](https://yaleits.atlassian.net/browse/YALB-1657)

- [x] Navigate to the [Interactive gallery component](https://deploy-preview-313--dev-component-library-twig.netlify.app/?path=/story/organisms-galleries--interactive-grid)

https://github.com/yalesites-org/component-library-twig/assets/366413/8cf25169-04f2-442b-84d4-2c0db659cb48

---

#### Testing: [YALB-1658 - Update search box styles](https://yaleits.atlassian.net/browse/YALB-1658)

- [x] Navigate to the [Utility nav component](https://deploy-preview-313--dev-component-library-twig.netlify.app/?path=/story/organisms-menu-utility-nav--utility-nav)

Screenshot 2023-12-12 at 10 48 48 AM

### Design Review
- [ ] YALB-1657 - Verify the designs match the [Gallery Caption Expand/Collapse updates - Heading plus text](https://www.figma.com/proto/4Lx9A5sxO08DxUNbewalrR/Yale-Sites---ideation?type=design&node-id=777-3587&t=JGXpbYjdrYShkJZB-0&scaling=min-zoom&page-id=750%3A5883&starting-point-node-id=777%3A3587&show-proto-sidebar=1) 
- [ ] YALB-1658 - Verify the designs match the [Search Figma Designs- Option C](https://www.figma.com/proto/4Lx9A5sxO08DxUNbewalrR/Yale-Sites---ideation?type=design&node-id=787-6112&t=JGXpbYjdrYShkJZB-0&scaling=min-zoom&page-id=644%3A3551&starting-point-node-id=787%3A6112)
  - note: the search follows header/global themes

### Accessibility Review


codechefmarc (approved):
This all worked great for me, approved. One possible question would be that when opening the caption, it is a smooth animation, but when closing it closes immediately. If this is intended, that's great, but wondered if it would look better if there was also a close animation too, even if it were faster than the open.

joetower:
@codechefmarc Thank you, Marc! 

I agree about animating the close behavior. It would be ideal, though I think it works fine as-is - users may appreciate the caption closing more quickly, if they even close it before moving to the next image. If we want to animate the close we'd need to do it in JavaScript, which I'd rather not do. 

In CSS, you [can animate the height of an element from, say a `max-height: 0` to `max-height: 500px` and the animation will work in both directions](https://css-tricks.com/using-css-transitions-auto-dimensions/). In our case, we want to animate from one dynamic height to another and back again.

Please let me know if you have any ideas.

joetower:
Actually, I think I have an idea. I'll take a look again tomorrow @codechefmarc 👍

kara-franco (changes requested):
Super nice work! The only item I found is to add the aria-expanded to the button so that screen reader users can hear the interaction. 

- [ ] Remove aria-expanded and aria label from `/html/body/div[5]/div[1]/div[2]/div/div/div[11]/div`
- [ ] Add aria-expanded to +/- button
- [ ] Update the JS to swap out True / False when the button opens and closes

kara-franco:
Screenshot 2023-12-13 at 12 17 02 PM
 
Screen shot of where to remove the ARIA

joetower:
@kara-franco Thank you! 
I should have mentioned this in the testing criteria, but now the entire caption text is clickable - you can expand/collapse the caption by clicking any part of the caption, not just the button. With that in mind, does that change your feedback? Do we need to adjust anything else here in light of changing the click event from the button to the entire text?

dblanken-yale (approved):
This is functioning very well.  Great work!  Something we may want to create a ticket for or discuss is whether those components should still allow a link in the CKE content portion.  Right now, I think they do.  Since clicking anywhere inside of the expanded version collapses, it makes me wonder if any links embedded will work still?

Screenshot 2023-12-14 at 7 37 26 AM

But as far as this goes, it's wonderful!

joetower:
@dblanken Oh, yeah, that's a great point! Maybe we shouldn't make the entire text clickable then. 🤔

joetower:
@kara-franco @dblanken-yale I update the expand/collapse toggle to apply _only_ to the actual button. The caption text is no longer clickable. Dave, that was an excellent point about allowing links in the captions. 

https://deploy-preview-313--dev-component-library-twig.netlify.app/?path=/story/organisms-galleries--interactive-grid

Please take a look and let me know if this is good to go. Thanks!

nJim (changes requested):
I'm requesting one style change. The plus-sign sometimes displays in the correct vertical position, while it occasionally shift down to lower in this space. I'm guessing it's related to the content in this space. Attaching a screenshot from my phone (iOS Safari) but I have also been able to recreate this on my Laptop in Chrome. This centering issue appears when the screen heigh it sufficiently small. Is it even an issue for our target audience? I'm usually a few generations behind in tech.
IMG_1003

On a related note, the pager slowly starts to hide when there is not enough vertical screen space. I think this is a separate problem to tackle and once again may not be a real issue if we are outside of the realistic viewport size. Still something to document and discuss.
IMG_1004

joetower:
@nJim That's a good call. I wondered if it was time to refactor some of the old code. I updated the button and caption alignment. Please give this another look, you should see the following: 

https://deploy-preview-313--dev-component-library-twig.netlify.app/?path=/story/organisms-galleries--interactive-grid

https://github.com/yalesites-org/component-library-twig/assets/366413/3f002c56-baff-446b-b2bb-0e83058b6690

---

For the pager, originally we had the pager hidden at smaller viewports. We'd need to look at refactoring the grid row definitions. It's a bit complicated because of the overflows. Anyway, I agree we should look at this in a separate ticket.

kara-franco (approved):
Looks and sounds great! Nice work!

PR #325, opened 2024-01-17
rel-100-mailto-shadow-fix: set text-shadow to inherit for text-with-link__text links

dblanken-yale:
## [[YALB-XX: Title](https://yaleits.atlassian.net/browse/YALB-XX)](rel-100-mailto-shadow-fix: set text-shadow to inherit for text-with-link__text links)

For text-with-image__text a elements, override the atoms.link mixin to not put a text shadow on links.

### Description of work
- Changes `.text-with-image__text a` elements to inherit their text-shadow, removing the extra shadow from CKE elements linked without proper data attribute decoration.

### Testing Link(s)

### Functional Review Steps

### Design Review

### Accessibility Review

### DIFF
--- a/components/02-molecules/text-with-image/_yds-text-with-image.scss
+++ b/components/02-molecules/text-with-image/_yds-text-with-image.scss
@@ -193,6 +193,8 @@ $component-content-spotlight-themes: map.deep-get(
 
   a {
     @include atoms.link;
+
+    text-shadow: inherit;
   }
 
   > *:last-child {


dblanken-yale:
@joetower Mike wanted me to double check with you on the ramifications of removing the text-shadow on these elements as they are causing an extra shadow on the Drupal side for CKE links.  You can see what it's doing here: https://github.com/yalesites-org/yalesites-project/blob/09ea402fd2f1addf1c5490da401cae3de7f6e0ce/web/profiles/custom/yalesites_profile/modules/custom/ys_starterkit/src/Plugin/Action/MediaBulkExport.php#L10

myvaughn:
If you add a link to the text of the spotlight and apply one of the component theme's you'll see the issue.
image

joetower:
@dblanken-yale  @myvaughn I see, thank you! 

Instead of adding `text-shadow: inherit` I would either: 
- add the text-shadow color variable so any `a` element in the component gets the background-color (which will change with the theme selection) instead of `inherit` so it would read: 

```  
a {
  @include atoms.link;
  
  text-shadow: var(--color-spotlight-text-shadow);
}
```

Or, maybe a better approach, if you're okay not having an animated underlines on any `a` element inside the component, would be to change the link to a plain-link like: 

```
a {
  @include atoms.plain-link;
}
```

This way the shadow won't appear at all for these links. 

---
Please take a look and let me know if you have any questions or if you'd like to discuss it further. Also, we should add this same fix to the portrait version.

dblanken-yale:
@joetower I'll play with both of those.  I'm kinda liking the plain-link idea initially, but want to see it.  Thanks for those suggestions!

I'll post pics of each here of what I find.

dblanken-yale:
As much as I love the plain-link, I feel we need to go with the shadow-text override for consistency (for now).  What do you all think?

dblanken-yale:
Publishing the text-shadow override version so we can look at it in Drupal also.  Once this is up I'll rebuild multidev and we can look.

joetower:
@dblanken-yale Okay, apologies, I gave you bad advice. 😭 .

So, instead of adding a text-shadow property at all to this block: 
```
  a {
    @include atoms.link;
  }
```
Don't do that. I forgot that this is set higher up in this file and we need to include this `a` element in the theme iteration. 

On line 49, if you make this code block match the following (add comma to .link, add the text-with-image__text class): 
```
      // override text-shadow color for links in component themes
      .link,
      .text-with-image__text {
        --color-text-shadow: var(--color-spotlight-text-shadow);
        --color-link-hover: var(--color-link-hover);
      }
```

This will properly re-map the `color-text-shadow` var from the link mixin to our color-spotlight-text-shadow which gets iterated over with theme. You should then see the following for default and theme options. 

https://github.com/yalesites-org/component-library-twig/assets/366413/27d795ce-a0fc-4e87-896d-e9b1aad137d4

dblanken-yale:
@joetower I think that did the trick.  If everyone could give it a look both in Drupal and Storybook, I think we can push this one through?

joetower (approved):
@dblanken-yale Looks good to me 👍 

Thanks for the collaboration on this one, Dave!

miketullo95 (approved):

PR #326, opened 2024-01-18
rel-100-spotlight-margin-start-fix: Fix padding in between spotlight padding

dblanken-yale:
## rel-100-spotlight-margin-start-fix: Fix padding in between spotlight padding

In the original, I believe we were attempting to target the block so that we can make margin-block-start 0, but in the case we were using, it was not selecting it.  I beleve this is correclty targeting now.

Mike T found that there was a gap between the two spotlights.

Screenshot 2024-01-18 at 11 59 07 AM

### Description of work
- Specifically target the element so we can apply a margin-block-start to 0
- Mirror this in both spotlight places just in case

### Testing Link(s)
- [ ] Navigate to the [Basic Spotlights](https://deploy-preview-326--dev-component-library-twig.netlify.app/?path=/story/page-examples-standard-pages--basic-spotlights)

### DIFF
--- a/components/02-molecules/content-spotlight-portrait/_yds-content-spotlight-portrait.scss
+++ b/components/02-molecules/content-spotlight-portrait/_yds-content-spotlight-portrait.scss
@@ -214,3 +214,8 @@ $component-content-spotlight-port-themes: map.deep-get(
     margin-bottom: 0;
   }
 }
+
+.content-spotlight-portrait:not([data-component-theme='default'])
+  + .content-spotlight-portrait__test:not([data-component-theme='default']) {
+  margin-block-start: 0;
+}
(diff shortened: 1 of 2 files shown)


joetower (commented):
@dblanken-yale ⁠I would move the logic you added. 

If you look at these lines: https://github.com/yalesites-org/component-library-twig/blob/1e4a1299f541fab1b122f096974421fee165c211/components/02-molecules/content-spotlight-portrait/_yds-content-spotlight-portrait.scss#L34C1-L36C6


I would add this to the selector stack instead: 


```
+ .content-spotlight-portrait:not([data-component-theme='default']),
+ .text-with-image:not([data-component-theme='default'])
```


Then the spacing would be removed from both instances. If a text-with-image not default theme follows a portrait, then it too should not have top margin. 


We'd want to do the same for the text-with-image


```
+ .text-with-image:not([data-component-theme='default']),
+ .content-spotlight-portrait:not([data-component-theme='default']),
```


I don't love this, but if I'm understanding the issue correctly, this should work for the top margin anyway.

dblanken-yale:
After our standup, it was decided to not include this in the release, and to create a ticket so we can address this in the future.  I added the ticket and referenced this PR.  Closing for now.

PR #330, opened 2024-01-24
YALB-889 - Component: Design Infographic Component - Storybook implementation

joetower:
### Description of work
- Adds `stat` component, with dials for `alignment`, `presentation style`, `font style`, and `component theme` so that it can be used on its own, outside of a `stats` component  
- Adds `stats` component, which can house up to 8 `stat` items. 
- The `stats` component can set `alignment`, `presentation style`, and `font style` for the whole group. Optionally, you can toggle on/off a `background image` and if all stat items should show `stat icons` (for Storybook only). The actual stats component in Drupal should govern `alignment`, `presentation style` and `font style` for all `stat` items housed within it. Not all stat items need to have icons, for example, as this will likely be on a stat-by-stat basis.

### Testing Link(s)
- [ ] Navigate to the [Stat Component Story](https://deploy-preview-330--dev-component-library-twig.netlify.app/?path=/story/molecules-stat--stat)
- [ ] Navigate to the [Stats Component Story](https://deploy-preview-330--dev-component-library-twig.netlify.app/?path=/story/organisms-stats--stats)

---

#### Testing the new `stat` component
- [x] Navigate to the `stat` component: https://deploy-preview-330--dev-component-library-twig.netlify.app/?path=/story/molecules-stat--stat
- [x] Verify you see three stat examples followed by a playground where the changes you make in Storybook's control pane will affect the stat shown. 
- [x] Verify you can set `alignment` `presentation style` and `font style`
- [x] Verify you can choose whether or not an icon should be shown
- [x] Verify you can choose a component theme
- [x] Test different combinations

https://github.com/yalesites-org/component-library-twig/assets/366413/ed82c4ef-c09a-48a8-b089-900cc9675241

---

#### Testing the new `stats` component
- [x] Navigate to the `stats` component: https://deploy-preview-330--dev-component-library-twig.netlify.app/?path=/story/organisms-stats--stats
- [x] Verify that you can choose: `alignment`, `presentation style`, `font style`
- [x] Verify you can set `show icons` and all stat items will display an icon
- [x] Verify you can choose a global theme and component theme
- [x] Test different combinations
- [ ] Test responsiveness

https://github.com/yalesites-org/component-library-twig/assets/366413/52a26d62-f3f0-471b-b01b-dc1771725b5a

---

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/proto/4Lx9A5sxO08DxUNbewalrR/Yale-Sites---ideation?type=design&node-id=914-5210&t=CeoRwLuCWLWxFPnr-0&scaling=min-zoom&page-id=851%3A3551)

### Accessibility Review


dblanken-yale (approved):
Awesome work as always.  I want to go through the sass file to see what else I can learn, but always neat to see you come up with this.

joetower:
@kara-franco Thank you! 

Technically, this is a standalone card https://deploy-preview-330--dev-component-library-twig.netlify.app/?path=/story/molecules-stat--stat

But, to make it look like a proper standalone card, it would still be in a Stats component just with one item. 

Would it make more sense, for Storybook, to have this Stat wrapped in the same markup as a Stats group? Or is it okay to visualize it that way it is by itself? 

For example: a Stat itself doesn't have padding. It only appears to have spacing when it's next to other Stat items. 

Perhaps a simple way of looking at it would be, in Drupal, would we have two components: 
- Stat - one single stat, as a card, no title, content, link, bg color, etc
- Stats - enter up to 8 stat items, has a title, content, link, bg color, etc

If someone were to use a single stat item, how would it layout against other content? If they wanted to use one stat, would they simply use the Stats component and only add one Stat?

codechefmarc (approved):
This looks very good. Approving with some comments, not necessary changes:

I had a similar comment to Kara with the left alignment of a singular card, but if it's always inside a stats container, then this is no big deal for the individual card. However, would be interesting to possible discuss having an individual stat card that could be used inline someplace else? That's a different topic though.

Another thought I had was about responsive when it gets small (phone sized) that the background image doesn't really look like anything - this could be the selected image, for example, but wondering (if it's decorative only?) to essentially hide the background image on mobile?

kara-franco (approved):
I totally get what what you're saying Joe! I think the designs (to me) were two separate componenets, Stat group (where up to 8 related stats can display) and Stat cards (where you can have 1-3 stats in squares). 

I can see a use case for Stat cards, especially in a right or left rail, to highlight a stat for a profile for example. We should explore this separately from this work!

Beautiful work!

joetower:
@kara-franco and @codechefmarc 

I gave your feedback some thought and added an optional data-attribute to the stats container. If `data-stat-collection-type="single"` is set on the Stats parent element, the stat (`.stat__inner` element) will have padding all around the item. 

For example: https://deploy-preview-330--dev-component-library-twig.netlify.app/?path=/story/molecules-stat--stat now has padding, but if the attribute isn't set to single (or present at all), as in a Stats component, that padding isn't there: https://deploy-preview-330--dev-component-library-twig.netlify.app/?path=/story/organisms-stats--stats&args=statsIcons:true

This way we never have to use `data-stat-collection-type` unless we do have single use case scenarios 👍 . At the very least, it makes the stat items nicer to look at in isolation. We _could_ update this in the future to accommodate additional collection-types. 

Thank you!

PR #332, opened 2024-02-02
YALB-1697: Bug: Margin for spotlights

joetower:
### Description of work
- Adds JS to scan the page for consecutive instances of `text-with-image` or `content-spotlight-portrait`
- If there are two consecutive instances, the first will have margin applied to the top and the second will have margin applied to the bottom.
- If there are several consecutive instances, the first will have top margin and the last will have bottom margin. 
- If a spotlight uses a default / no color theme option, it will have top and bottom margin, even if it is the only spotlight instance
- Spotlights, by default, do not have margins. Instead, we introduce margins if specific attributes are present.
- Margins are introduced if: 
  - the spotlight does not use a theme color (is set to `default`)
  - the spotlight is the only spotlight on the page
  - there are consecutive spotlights on the page, the first has top margin, the last has bottom margin

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-332--dev-component-library-twig.netlify.app/?path=/story/page-examples-standard-pages--basic-spotlights)

### Functional Review Steps
- [ ] Visit example page here: https://deploy-preview-332--dev-component-library-twig.netlify.app/?path=/story/page-examples-standard-pages--basic-spotlights
- [ ] Test on the Drupal multidev here: https://pr-563-yalesites-platform.pantheonsite.io/spotlights
- [x] Confirm that the first and last spotlight in each group (two or more spotlights) have margins applied
- [x] If first, there is top margin (`data-spotlights-position="first"`)
- [x] If last, there is bottom margin ( `data-spotlights-position="last"` _except if last component on the page and not using a component theme_)

https://github.com/yalesites-org/component-library-twig/assets/366413/726f037a-2a58-407f-9c78-f9a97ea4f605

- [x] Note that if either Content Spotlights, landscape or portrait, are using a theme and they are the last component before the footer, there is no bottom margin. 
- [x] Note that if you add a Content Spotlight set to Theme: Default - No Theme, there _is_ space between the component and the footer. 
- [ ] With one spotlight : https://pr-563-yalesites-platform.pantheonsite.io/one-spotlight-only
- [x] Note: if a spotlight is the one and only spotlight on the page, then it should have top and bottom margin (`data-spotlights-position="first-and-last"`)

https://github.com/yalesites-org/component-library-twig/assets/366413/fd8ad2aa-a9f7-41dd-8bb0-cb92126badfc


dblanken-yale (approved):
Works great!  Nice work!  I was going to suggest maybe combining the if statements at the end so that it doesn't overwrite itself in the even it is first and last, but then again that means there's only one spotlight and that's a trivial performance gain.  LOL

I really like the idea of creating the data attributes for this!  So easy to understand in the markup.

sskalisky-4k:
This looks great, @joetower!

miketullo95 (approved):
this is great!

codechefmarc (approved):
Whoops, reviewed on Atomic as well, I think that's correct. Anyways, approved! Nice work!

PR #339, opened 2024-02-26
YSP-323: breadcrumb overlayed into text block with heading when page title hidden

dblanken-yale:
## [YSP-323: breadcrumb overlayed into text block with heading when page title hidden](https://yaleits.atlassian.net/browse/YSP-323)

The issue was when hiding the page title, yet creating a block that contained a header for its first element, it would cause the header item to bleed into the breadcrumbs.  This was due to the visually hidden happening on the whole div.  In normal cases, the heading has a margin-top of 1em due to text-field-header-spacing on the h1.  We emulate this by targeting the div that still is present containing the visually hidden page title so that we still have the same spacing.

### Description of work
- Add margin-tops to both visually-hidden page titles and hidden page titles
- Add a storybook variable to display these in a standard page

### Testing Link(s)
- [ ] [Storybook](https://deploy-preview-339--dev-component-library-twig.netlify.app/?path=/story/page-examples-standard-pages--basic)

### Functional Review Steps
- [ ] Change the new `Page Title Display` select between each option and ensure that breadcrumbs don't bleed into the heading below it.
- [ ] Visit temporarily created multidev in [Yalesites PR](https://github.com/yalesites-org/yalesites-project/pull/584) and follow the steps there

### Accessibility Review

### DIFF
--- a/components/02-molecules/page-title/_yds-page-title.scss
+++ b/components/02-molecules/page-title/_yds-page-title.scss
@@ -26,7 +26,17 @@
 // Add margin to the following page-title display types:
 //   - visually-hidden
 //   - hidden (when no page-title div exists after the breadcrumbs__wrapper
+// Margins borrowed from
+// components/04-page-layouts/page-layouts.scss:page_layouts.scss:31-39
 div.page-title.visually-hidden + div,
 div.breadcrumbs__wrapper + :not(.page-title) {
-  margin-top: 1em;
+  margin-top: var(--size-spacing-10); // 4rem
+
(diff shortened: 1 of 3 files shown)


codechefmarc:
> In testing, I found a bug for the page title hiding. I'll put in a separate ticket for that. But, if you set the page title to be visually hidden, it removes it completely just like not displaying the title at all. In visually hidden, it should still be in the DOM.

Never-mind! I found it, was looking in the wrong spot.

miketullo95 (approved):
👍

PR #355, opened 2024-04-11
YSP-461: Localist: Style Event Meta Block with new designs

joetower:
### Description of work
- Adds localist version of the event meta block
- Wires new variables coming from Localist to new event-meta-localist component in the component library
- Removes JS from cards, making them pseudo fully-clickable in favor of more click-standard link behaviors. right-clicking over the image or the title will now work, without issue in any browser. 
- Main PR: https://github.com/yalesites-org/yalesites-project/pull/631
- Atomic: https://github.com/yalesites-org/atomic/pull/221

### Testing Link(s)
- [x] Mutlidev test event: https://pr-631-yalesites-platform.pantheonsite.io/events/katys-event
- [x] You can also test this in the main PR here: https://github.com/yalesites-org/yalesites-project/pull/631

#### Testing full event / [YSP-461: Localist: Style Event Meta Block with new designs](https://yaleits.atlassian.net/browse/YSP-461)
- [x] Navigate to: https://deploy-preview-355--dev-component-library-twig.netlify.app/?path=/story/molecules-meta--event-localist
- [x] Verify it resembles the design
  - [x] Note: the `more dates` presentation differs from the design in that they span the content-width instead of only spanning the left column. I'm in favor of the way it is in this branch opposed to the design, but I'm open to other's thoughts. I implemented it this way mostly due to how the existing components are structured, but also because an event with an image would likely push the `more dates` button element below the image anyway, in which case, it likely makes sense for the more dates to be full-width.

https://github.com/yalesites-org/component-library-twig/assets/366413/7e5aa97d-e89f-4e4f-8819-c3849fc603cf

---

#### Testing cards and [YSP-446: Cards Block - Right Click function using MS Edge](https://yaleits.atlassian.net/browse/YSP-446)

- [x] Navigate to: https://deploy-preview-355--dev-component-library-twig.netlify.app/?path=/story/molecules-cards--event-card
- [x] Test the `card`, `list`, and `single` collection types
- [x] Verify you see primary and secondary CTA buttons on all three collection types
- [x] Note that you can right-click over the `image` or the `title` and choose "open in new window/tab". Note that the full card is no longer clickable.
- [x] Change the collection type to `condensed`, verify the buttons do not appear
- [x] Note that in the condensed display, if an event is a multi-day event, a label renders after the date (multi-day event)
- [x] Toggle the Multi day event control to see how it is reflected in the cards (`list` and `condensed` only)

https://github.com/yalesites-org/component-library-twig/assets/366413/fa46832f-392c-4cd6-9e20-faa9afadfd11

### Design Review
- [x] Verify the designs match the [Figma Designs](https://www.figma.com/proto/4Lx9A5sxO08DxUNbewalrR/Yale-Sites---ideation?type=design&node-id=1669-2189&t=BqWZo0F54aKLuPcW-0&scaling=min-zoom&page-id=1569%3A2)

### Accessibility Review
- [ ] Note: the cards now have a link wrapped around the `image` (with an `aria-label` set to the content heading/title and with `tabindex= -1` `aria-hidden=true` attributes)
- [ ] The title/heading is still wrapped in a link. The entire cards still has a hover effect, but only the image and title are clickable or right-clickable (manually open in new tab, etc.).


codechefmarc (approved):
This looks great! You noted that the "multi-day" language only applies to list and grid, but it also applies to single. Perhaps it should also apply to grid as well?

dblanken-yale (changes requested):
For the Event Localist in Storybook, the "All Day" doesn't seem to change anything that I noticed.  Am I missing it?

[Current URL with options used](https://deploy-preview-355--dev-component-library-twig.netlify.app/?path=/story/molecules-meta--event-localist&args=withImage:false;format[0]:In-person)

dblanken-yale:
Ok the event card is just so good.  Nice work!  No issues that I saw there.

codechefmarc:
> For the Event Localist in Storybook, the "All Day" doesn't seem to change anything that I noticed. Am I missing it?
> 
> [Current URL with options used](https://deploy-preview-355--dev-component-library-twig.netlify.app/?path=/story/molecules-meta--event-localist&args=withImage:false;format%5B0%5D:In-person)

@dblanken-yale - I don't think the new design had anything that dealt with all day changing anything. The old design did, IIRC. For now, I think like other components, we'll leave this in as a control, but only add it to the front-end if/when we need to.

PR #365, opened 2024-05-14
fix(YSP-527): V1.3.0: Reference Card link color contrast

dblanken-yale:
## [YSP-527: V1.3.0: Reference Card link color contrast](https://yaleits.atlassian.net/browse/YSP-527)

In single card types like reference cards, the hover link and link would change color, which does not match Storybook, which keeps the text white, even when visited.  This attempts to get us closer to a consistent look with Storybook.

### Description of work
- Modifies what `--color-link-hover` is in the case of the heading link on a reference card to match color-link-hover
  - This mimics what grand hero is doing in a similar case

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-365--dev-component-library-twig.netlify.app/?path=/story/molecules-cards--profile-card&args=collectionType:single)

### Functional Review Steps
- Verify these in Safari, Firefox, and a Chromium-based browser as they had different behaviors
  - [ ] Verify the link text is using an overloaded `--color-link-hover` of `var(--color-link-hover)`
  - [ ] Visit the [Drupal Multidev](https://github.com/yalesites-org/yalesites-project/pull/651) to test this in Drupal as well
  - [ ] Ensure behavior is consistent cross browser

### Design Review

### Accessibility Review

### DIFF
--- a/components/02-molecules/cards/reference-card/_yds-reference-card.scss
+++ b/components/02-molecules/cards/reference-card/_yds-reference-card.scss
@@ -42,15 +42,6 @@ $break-single-card-collection-max: tokens.$break-xl - 0.05;
     --color-heading: var(--color-card-single-text);
 
     margin-block-start: var(--size-spacing-11);
-
-    // Overload the color to be the single text color to fix contrast issues.
-    & a.reference-card__heading-link {
-      color: var(--color-card-single-text);
-
-      &:visited {
-        color: var(--color-card-single-text);
-      }
-    }
   }
 
   // Condensed card collection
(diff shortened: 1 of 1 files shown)


miketullo95 (approved):
can confirm that the text stays white 😄

joetower:
@dblanken-yale I also can confirm it stays white. My only question is, do we want the `:visited` state to use our visited-link colors? 

I'm wondering if we want to refine this to account for the visited link variables (`var(--color-link-visited-base` and `var(--color-link-visited-hover)`) as we've done here: https://deploy-preview-365--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--grand-hero-banner

Screenshot 2024-05-14 at 14 30 32

kara-franco (approved):
Looks great, color contrast issue is gone. I agree with Joe, let's keep things consistent and add the visited state color (light purple)

dblanken-yale:
I attempted to mimic grand hero where it overloads the hover variable used.  It's still white as it's defined just above it, but I believe has the same pattern.

dblanken-yale:
Also, check it out on both Safari and a Chromium based.  On Safari, it was behaving, but in Chrome/Edge it was doing the blue text on hover.

miketullo95 (approved):
looks great! 👍

dblanken-yale:
Talked with @kara-franco and with @miketullo95 's blessing, she's fine with the change and we can address it later with the underline in Safari.

PR #367, opened 2024-05-17
YSP-509: Add Code Style to WYSIWYG

laura-johnson:
### Description of work
- Adds the ‘Code’ and ‘Code Block’ buttons to the CKEditor toolbar for the ‘Basic HTML’ text formatter with supporting styles
- Adds a demo of the code block in Storybook in the text molecule component.
- Adds highlight.js to the codeblock to highlight the code.

### Functional testing steps:
- [ ] Edit a text block in layout builder that uses the Basic HTML format, add a code snippet or block of code. For the block of code the default is plain text, there is a pull-down arrow on the right of that button that you can use to select from a number of code styles
- [ ] Verify they look as expected (current style shown below)
- [ ] Go to the text molecule in Storybook, verify you see a styled code block at the bottom

highlight-codeblock

Depends on: 
https://github.com/yalesites-org/yalesites-project/pull/655
https://github.com/yalesites-org/atomic/pull/234

### Notes
- I tried to import the CSS from the highlightjs library but webpack doesn't like that. So I grabbed the CSS and put it into the component SCSS. I'm not sure if there is a better way.
- There are a number of CSS styles to choose from for the highlighting. I chose one that I saw in a demo that looked good called "monokai-sublime". Not sure if the group would like to play around with the styles, you can do that here: https://highlightjs.org/demo
- I took my best guess at the different languages we might want. There are many that we could optionally include.


codechefmarc (approved):
All of this looks good, but would love to get Joe's eyes on it as this is more in his realm. I tested the new styling and it looks great! There is one small change, but not a blocker. There's an extra semicolon in the `typography.stories.js` at the end that shows up in Storybook. I don't think that was you, but while we've got this PR open, may as well remove that:
Screenshot 2024-05-24 at 1 58 24 PM

components/01-atoms/typography/text/_yds-text.scss
```diff
@@ -6,6 +6,95 @@ p {
...
+
+    /*
+    Monokai Sublime style. Derived from Monokai by noformnocontent http://nn.mit-license.org/
+    */
+
+    &.hljs {
+      background: #23241f;
+      color: #f8f8f2;
+      overflow-x: auto;
+    }
```
joetower, line 26: @laura-johnson I'd like to see all of these hex colors replaced with CSS variables or color tokens, if possible. Adding the colors as tokens is likely more involved than what's needed here, so maybe we could define the colors as CSS variables at the top of this file. 

Something like: 
```
// Code highlighting color variables

:root {
--color-hljs-text: #f8f8f2;
--color-hljs-background: #23241f;
}
```

And then, this part: 
```
   &.hljs {
      background: #23241f;
      color: #f8f8f2;
      overflow-x: auto;
    }
```

Would turn into: 

```
   &.hljs {
      background: var(--color-hljs-text);
      color: var(--color-hljs-background);
      overflow-x: auto;
    }
```

Please let me know if you'd rather tag me in to make these hex to CSS variable changes. 👍

components/01-atoms/typography/text/_yds-text.scss
```diff
@@ -6,6 +6,95 @@ p {
   p:last-child {
     margin-bottom: 0;
   }
+
+  // Code blocks.
+  pre code {
+    display: block;
+    padding: 1em;
```
joetower, line 13: @laura-johnson This `1em` padding should use a spacing variable. Could you please change this from `padding: 1em` to `padding: var(--size-spacing-5)`?

joetower (changes requested):

components/01-atoms/typography/text/_yds-text.scss
laura-johnson: Ok great, thanks Joe! I'll make these changes.

laura-johnson:
> All of this looks good, but would love to get Joe's eyes on it as this is more in his realm. I tested the new styling and it looks great! There is one small change, but not a blocker. There's an extra semicolon in the `typography.stories.js` at the end that shows up in Storybook. I don't think that was you, but while we've got this PR open, may as well remove that: Screenshot 2024-05-24 at 1 58 24 PM

I'll fix that semicolon, thanks @codechefmarc!

laura-johnson:
@codechefmarc @joetower I revised the SCSS as per Joe's comments and also added all of the languages requested by the team. And I removed the errant semicolon!

joetower (approved):
@laura-johnson The updated look great! Thank you!

PR #370, opened 2024-05-24
YSP-431 Wire/Add/Tweak Call-to-Action button utility nav

laura-johnson:
### Description of work
- Adds the CTA button to the header if the CTA fields are populated in the header form

### Functional testing steps:
- [ ] Go to /admin/yalesites/header, verify there is a new Call To Action section with Link Text (text) and Link Target (linkit) fields
- [ ] Add values for either or both fields
- [ ] Verify if both fields are populated, the Call to Action button appears in the upper right of the navigation as per the designs

Requires https://github.com/yalesites-org/yalesites-project/pull/665


nJim (approved):
Approved in weekly meeting with Kara and Mike T.

PR #377, opened 2024-06-14
YSP-541: Embed: Add Microsoft Forms

dblanken-yale:
### Other work related to this PR
- [Yalesites Project: PR #662](https://github.com/yalesites-org/yalesites-project/pull/662)

### Description of work
- Adds MS forms embed example to storybook
- Updated forms embedType height to be `90vh`

### Testing Link(s)
- [ ] Navigate to the [MS Forms Component Story](https://deploy-preview-377--dev-component-library-twig.netlify.app/?path=/story/molecules-embed--embed-microsoft-forms)

### Functional Review Steps
- [ ] Verify you can see the example MS Form embed
- [ ] Verify that the height of the form and possibly other embeds like PowerBI and Qualtrics are acceptable

### Design Review

### Accessibility Review

### DIFF
--- a/components/02-molecules/embed/_yds-embed.scss
+++ b/components/02-molecules/embed/_yds-embed.scss
@@ -10,7 +10,7 @@
 }
 
 $heights: (
-  'form': 120vh,
+  'form': 90vh,
   'audio': 130px,
   'unknown': 70vh,
 );
--- a/components/02-molecules/embed/_yds-embed.scss
+++ b/components/02-molecules/embed/_yds-embed.scss
@@ -10,7 +10,7 @@
 }
 
 $heights: (
-  'form': 70vh,
+  'form': 120vh,
   'audio': 130px,
   'unknown': 70vh,
 );
--- a/components/02-molecules/embed/embed.stories.js
+++ b/components/02-molecules/embed/embed.stories.js
@@ -70,3 +70,13 @@ export const EmbedPowerBI = ({ width, type, loading }) =>
     embed__loading: loading,
     embed__type: type,
   });
+
+export const EmbedMicrosoftForms = ({ width, type, loading }) =>
+  embedTwig({
+    embed__src:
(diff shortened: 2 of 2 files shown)


kara-franco (changes requested):
Is there a way to add text to the iframe title attribute. it's empty and throwing an a11y error.

miketullo95:
my only note is that we have gotten feedback from the qualtrics embed that users want the iframe to be longer and show more q's without scrolling. Idk if this is blocking approvals, but this feedback could be mirrored with this

dblanken-yale:
Ya we're using the same embed type for it.  Want me to elongate it which would do the same for qualtrics?  Then you can see how it'd be and we could incorporate that change here. @miketullo95

dblanken-yale:
Maybe from 70vh to like 120vh?

dblanken-yale:
Building now; check it out at 120vh.

miketullo95:
> Maybe from 70vh to like 120vh?

yeah lets try 120! this would essentially close https://yaleits.atlassian.net/browse/YSP-343 😄

dblanken-yale:
Yalesites project version is also building to show this off now.

dblanken-yale:
Ok all built in both places; you should be able to see this vh change in any embedType of forms (PowerBI, etc.)

miketullo95:
what do you think about the new height @kara-franco? my initial thought is maybe too long and maybe 90 is a good middle-ground. But looking at it from a "how many questions fit before scrolling" perspective, it moves it from like ~2.5 to ~4.5 when looking at qualtrics. 

for comparison:
new -  https://pr-662-yalesites-platform.pantheonsite.io/
old - https://dev-yalesites-mt927.pantheonsite.io/ (scroll down)

dblanken-yale:
The nice thing is this is an easy change I think so we can try 90 if you both think that's worth trying.

miketullo95:
in the essence of getting this done by noon, i think lets go with 90 and see if users still want a taller form embed haha

dblanken-yale:
Sounds good; incoming!!!!!

nJim (approved):

miketullo95 (approved):

PR #393, opened 2024-07-31
YSP-583: Event date display a range

chaohlu:
### Description of work
- Updated start and end date formatted
- Updated start and end time formatted
- Delete comma and change "to" to "-" 

### Testing Link(s)
- [ ] Navigate to the [Localist Event](https://deploy-preview-393--dev-component-library-twig.netlify.app/?path=/story/molecules-meta--event-localist)
- [ ] Navigate to the [Event Page](https://deploy-preview-393--dev-component-library-twig.netlify.app/?path=/story/molecules-meta--event)

### Functional Review Steps
- [ ] Visit storybook Localist Event link
- [ ] Check the date field should look format example: Wed July, 31 2024 12:00pm-2:00pm
- [ ] Test to see the changes will work on different date and time.
- [ ] Test the above using the Event Page also
- [ ] Test this functionality inside of the [multidev](https://github.com/yalesites-org/yalesites-project/pull/717)

### UX
- [ ] Is the new format acceptable

### Accessibility Review

### DIFF
--- a/components/02-molecules/meta/event-meta/yds-event-meta-localist.twig
+++ b/components/02-molecules/meta/event-meta/yds-event-meta-localist.twig
@@ -180,6 +180,8 @@
             <ul {{ bem('multiple-dates', [], event_meta__base_class) }} aria-expanded="false">
               {% for event_date in event_dates %}
                 <li {{ bem('multiple-dates__date', [], event_meta__base_class) }}>
+                {{ event_date.formatted_start_date }} {{ event_date.formatted_start_time }} to {{ event_date.formatted_end_date }} {{ event_date.formatted_end_time }}
+                </li>
               {% endfor %}
             </ul>
           </div>
--- a/components/02-molecules/meta/event-meta/yds-event-meta-localist.twig
+++ b/components/02-molecules/meta/event-meta/yds-event-meta-localist.twig
@@ -138,9 +138,9 @@
         <div {{ bem('date', [], event_meta__base_class) }}>
 					{% if event_dates.0.formatted_start_date == event_dates.0.formatted_end_date %}
 						{{event_dates.0.formatted_start_date| date('D M j, Y')}}
-						{{ event_dates.0.formatted_start_time| date('g:i a') }}
+						{{ event_dates.0.formatted_start_time| date('g:ia') }}
 						-
-						{{ event_dates.0.formatted_end_time| date('g:i a') }}
+						{{ event_dates.0.formatted_end_time| date('g:ia') }}
 					{% else %}
           	{{ event_dates.0.formatted_start_date }}
 						{{ event_dates.0.formatted_start_time }}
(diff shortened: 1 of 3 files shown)


components/02-molecules/meta/event-meta/yds-event-meta-localist.twig
```diff
@@ -168,8 +181,7 @@
             <ul {{ bem('multiple-dates', [], event_meta__base_class) }} aria-expanded="false">
               {% for event_date in event_dates %}
                 <li {{ bem('multiple-dates__date', [], event_meta__base_class) }}>
-                  {{ event_date.formatted_start_date }} {{ event_date.formatted_start_time }} to {{ event_date.formatted_end_date }} {{ event_date.formatted_end_time }}
-                </li>
+          {{ event_dates.0.formatted_start_date }} {{ event_dates.0.formatted_start_time }} - {{ event_dates.0.formatted_end_time }}                 </li>
```
dblanken-yale, line 184: I believe this needs to mimic the above if statement you created also so that they are displaying the same.  What do you think?

Also, you are referencing `event_dates` here when you're looping through them, so you'd want `event_date` I believe.  Looking at the changed code you'd just reference them like this:  `event_date.formatted_start_date`

dblanken-yale (changes requested):

components/02-molecules/meta/event-meta/yds-event-meta-localist.twig
dblanken-yale: So while you deleted your code in your latest push, you didn't revert to the old code that was there.  ☹️ You'll want to place back the above that's in red if the green wasn't meant to be there.

components/02-molecules/meta/event-meta/yds-event-meta-localist.twig
```diff
@@ -168,8 +180,6 @@
             <ul {{ bem('multiple-dates', [], event_meta__base_class) }} aria-expanded="false">
               {% for event_date in event_dates %}
                 <li {{ bem('multiple-dates__date', [], event_meta__base_class) }}>
-                  {{ event_date.formatted_start_date }} {{ event_date.formatted_start_time }} to {{ event_date.formatted_end_date }} {{ event_date.formatted_end_time }}
```
dblanken-yale, line 171: So while you deleted your code in your latest push, you didn't revert to the old code that was there. ☹️ You'll want to place back the above that's in red if the green wasn't meant to be there.  See the last resolved version above to see that change.

dblanken-yale (changes requested):

components/02-molecules/meta/event-meta/yds-event-meta-localist.twig
chaohlu: So I have to put the original code back push to git and then fix it?

dblanken-yale: Either the original code back or make appropriate changes that should go there instead.  You can do it in sequential order, too.  Like first restore the original code, then make another commit with your new change.  That one is up to you.  :D

chaohlu: I already put the original code on line 183 back. Didnt make any new changes on 183 until issue resolve.

dblanken-yale (approved):
These changes match what was requested.  I did have questions to UX to verify that the time formats are what is expected since there was specific logic in the past for not showing `:00` if it was there.  But this code ensures that both the localist and regular event dates are the same.

Throwing out there that a possible refactor of this in the future might be to have the localist twig file utilize the yds-date-time and reuse the format so that special logic doesn't have to go there.  I'd think all dates would use this partial.  But outside the scope of this ticket.

Codewise this makes sense.

dblanken-yale:
@chaohlu It's customary to wait till you get approvals from more than just code, so hold off merging until then.  Also, I'm going to make a multidev like we talked about showing this off this time, and we can get together later to go over how to do this on your own.

atiddei (approved):

kara-franco (approved):

PR #394, opened 2024-07-31
YSP-604: Grand Hero Video doesn't load on mobile

dblanken-yale:
### Description of work
- Adds `playsinline` to video tag in adherence to Safari's newer policies

### Testing Link(s)
- [ ] Navigate to the [Grand Hero with video enabled](https://deploy-preview-394--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--grand-hero-banner&args=withVideo:true)

### Functional Review Steps
- [ ] Verify that on mobile, the video autoplays
- [ ] Verify that on desktop Safari, the video autoplays
  - Keep in mind you must not be in low power mode for this to be successful
- [ ] Try in other browsers to ensure they still work
- [ ] [Test it in Drupal](https://pr-714-yalesites-platform.pantheonsite.io/)
  - Video should be on grand hero on front page

### DIFF
--- a/components/01-atoms/videos/video-background/yds-video-background.twig
+++ b/components/01-atoms/videos/video-background/yds-video-background.twig
@@ -14,7 +14,7 @@
 
 <div {{ bem(video_background__base_class, video_background__modifiers, video_background__blockname) }} data-component-theme={{ video_background__button__background_color }}>
   {% block video_background__content %}
-    <video muted>
+    <video muted playsinline>
       <source src="{{ video_background__content }}" type="video/mp4">
     </video>
   {% endblock %}


vinmassaro:
In iOS Safari and Firefox, the video does not autoplay on load for me, but does play inline now when I click the play button. On the Drupal example, I see a black background until I tap the play button. While it does not follow the functional review steps, I would prefer this behavior on smaller screen sizes because loading a 25MB video up front over a cellular connection is a bad experience and a lot of data to transfer, and the user should have a choice in downloading it. We should probably implement a [static image](https://til.simonwillison.net/html/video-preload-none), though.

dblanken-yale:
@vinmassaro What version of iOS are you currently running?  And just verifying because mine starts each time, is reduced motion turned off?  It seems to be working in browser stack.  ☹️ I agree with a static image.  Just want to make sure we understand what is making our experiences different.

vinmassaro:
@dblanken-yale yes, reduce motion is enabled. It now autoloads for me if I toggle that setting 👍

kara-franco (approved):
Nice, the video autoplayed for me in all Mac browsers and in Android browser. 

There was a very slight delay in Safari, probably Safari being Safari.

PR #415, opened 2024-09-10
YSP-666: Breadcrumbs not clickable in Firefox when no banner blocks provided in mega menu mode

dblanken-yale:
## [YSP-666: Breadcrumbs not clickable in Firefox when no banner blocks provided in mega menu mode](https://yaleits.atlassian.net/browse/YSP-666)

By changing the visibility from hidden to a display of none, this allows the links to show through for Firefox browsers to be clickable.  I'm worried about the A11y implications of this, as I assume visibility was used on purpose.

### Description of work
- Change mega-menu submenu items when not shown on page from using `visibility` to `display none` so breadcrumb links can be clickable.

### Testing Link(s)
- [ ] Navigate to the [Menu Story](https://deploy-preview-415--dev-component-library-twig.netlify.app/?path=/story/organisms-menu-primary-nav--primary-nav&args=menuVariation:mega)

### Functional Review Steps
- [ ] Verify you can use the menu and it still looks the same

### Accessibility Review


dblanken-yale:
@kara-franco I wanted you especially to take a look at this one.  The mega menu subitems seem to be somehow interfering with the breadcrumbs only for Firefox when in a specific scenario.  Do you have any background why we might use `visibility: hidden` for an `aria-expanded` of false?  While this does seem to fix the issue for Firefox, my worry is that we're breaking something else.  I wanted to double check.

I'd start a multidev with it but we're out at the moment.  I could show you locally though, or you can visit the problem page in the ticket and make the one-line change.

kara-franco (approved):
Looks good! I actually cannot tell you why that visibility hidden was there.

PR #435, opened 2024-11-20
YALB-263: Show CAS'd content in search results | CL | Heading Prefix Icon -> search result prefix icon

joetower:
### Description of work
- Adds heading_icon_prefix to allow for embedding an icon before a heading
- Includes the new `heading__prefix__icon` as `search_result__prefix__icon` in `02-molecules/search-result/yds-search-result.twig`

### Testing Link(s)
- [ ] Test here: https://github.com/yalesites-org/yalesites-project/pull/800
- [x] In component library: navigate here: https://deploy-preview-435--dev-component-library-twig.netlify.app/?path=/story/molecules-search-result--search-result&args=contentType:Page;isCas:!true
- [x] Verify that when you toggle `is_cas` to `true` the search result snippet disappears and a lock icon appears

[IYY-search-result-lock.webm](https://github.com/user-attachments/assets/e13c2c7c-3c69-4939-8874-9077e9c68d06)


joetower:
@dblanken @codechefmarc , the CL side of this ticket is ready to review here and as part of the main project PR https://github.com/yalesites-org/yalesites-project/pull/800

Dave, once I ran the `local:git-checkout`, then checked out all the correct branches (`IYY-263--cas-search`) I was getting a Storybook broken build due to these lines: 
```
// Global link treatment
// import '../lib/link-treatment/link-treatment.js';
// import 'linkpurpose/css/linkpurpose.css';
// import '../lib/link-treatment/link-treatment.scss';
```

After commenting them out, the build works fine again. I'm not sure the best way to get these to play nicely with our Storybook/Webpack setup. Do you have any ideas? Also, :wave: again.

dblanken-yale:
@joetower Welcome back!!!  😁

Yes, that is strange; builds aren't bombing and it's compiling for me locally.  That is a trickier one in that I had to add a loader to handle it since linkpurpose was a POJO class.  I assume you did, but maybe do a `npm i -D` to make sure the loader is downloaded.  Maybe we can either meet up or you can give me the full output so I can look into it more.

I'm going to bring this branch down too to see what could be happening.

dblanken-yale:
@joetower Could you be using Dart Sass 3.0.0+?  I noticed this message in my compilation, but it still built:

```
WARN Deprecation Warning on line 0, column 8 of file:///Users/db2553/code/yalesites-project/web/themes/contrib/atomic/_yale-packages/component-library-twig/lib/link-treatment/link-treatment.scss:0:8:
WARN Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.
WARN
WARN More info and automated migrator: https://sass-lang.com/d/import
WARN
WARN 0 | @import '../../node_modules/linkpurpose/css/linkpurpose';
WARN
WARN
WARN lib/link-treatment/link-treatment.scss 1:9  root stylesheet
```

dblanken-yale:
Migrating to `@use` from `@import` squelches that warning for me; if you're using a later Dart, maybe that'll let it build?

dblanken-yale (approved):
This is so good.  <3

Welcome back @joetower and great job as always!!!

dblanken-yale:
Adding @ntay90 and @miketullo95 for their input, but excited to see this.

joetower:
@dblanken Thank you! We should update all of those imports, but that should probably be in a new tech debt ticket. 

I was incorrect about the cause of the build issues I was seeing. Apologies for calling out the link-treatment work as a possible cause. I sent you a message in Teams, as well. I'll include here what I wrote there, as well. 

So, I didn't express the issue I was seeing very well in my PR comment. I think what I was experiencing could be something caused by node/storybook/package updates.  When I was looking at the live build of the Storybook instance to refresh my memory on the animation work we've done Storybook but not implemented in Drupal, I noticed the config page broken https://yalesites-org.github.io/component-library-twig/?path=/story/config--global-config so I couldn't test the animation work (which may be unrelated to what I'm stream-of-thought outlining below). Locally, I don't see this issue, but I was experiencing other oddities which may be tech debt after node package updates or something else that has changed. 
 
Thankfully, I think Storybook is compiling okay and what I was seeing wasn't a blocker. We just have some tech debt to figure out. There is an issue with the local storage get/put setup in config and Storybook right now. The config settings aren't saving and persisting, which could be related to Storybook's "Update Storybook" functionality? Maybe we just need to run a migration https://storybook.js.org/docs/api/csf ? Looks like we may need to update our stories files too.
 
Anyway, none of that is a blocker. When we look at adding more motion work, we will likely want to address the tech debt. 

---

I'll be curious to know if we need to do anything to help this CAS'd content be expressive in terms of accessibility, @ntay90 . Do we need to add aria attributes so screenreaders are aware that some content is hidden because you need to login to read the full content?

miketullo95 (approved):

PR #437, opened 2024-11-21
YSP-752: Storybook global config settings don't use localStorage any longer

dblanken-yale:
### Description of work
- Uses `args` in config story
- Create way to use localStorage with a default fallback on values
- Remove "Save" feature when updating stories
- Use `@use` over `@import` for `link-treatment.scss`
- Remove `console.log` message
- Defined `resetAttributes` to the window object so we don't get redefinition errors

### Testing Link(s)
- [ ] Navigate to the [Global Config](https://deploy-preview-437--dev-component-library-twig.netlify.app/?path=/story/config--global-config)

### Functional Review Steps
- [ ] Verify that defaults appear in `Controls` section
- [ ] Verify that upon updating items that use localStorage, that they persist when you visit again
- [ ] Verify that on updating a control item, the "Save/Update" question at the bottom no longer shows
- [ ] Use resetAttributes and make changes noting that there are no console errors on resetAttributes already being defined

### DIFF
--- a/components/_settings/config.stories.js
+++ b/components/_settings/config.stories.js
@@ -44,102 +44,113 @@ export default {
       name: 'Site: Animation Theme',
       options: siteAnimationOptions,
       type: 'select',
-      defaultValue: 'default',
     },
     thickness: {
       name: 'Site: Line thickness',
       options: thicknessOptions,
       type: 'select',
-      defaultValue: 'hairline',
     },
     dividerColor: {
       name: 'Site: Line color',
       options: ['gray-500', 'blue-yale', 'basic-brown-gray'],
       type: 'select',
-      defaultValue: 'gray-500',
     },
     dividerWidth: {
       name: 'Site: Divider width',
       options: [...widths],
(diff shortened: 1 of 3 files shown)


components/_settings/config.stories.js
```diff
@@ -34,6 +34,110 @@ const themesOneToEight = [
...
+  siteHeaderImage: getValueFor('siteHeaderImage'),
+  siteHeaderSiteNameImage: getValueFor('siteHeaderSiteNameImage'),
+  headerBorderThickness: getValueFor('headerBorderThickness'),
+  siteFooterVariation: getValueFor('siteFooterVariation'),
+  siteFooterTheme: getValueFor('siteFooterTheme'),
+  siteFooterAccent: getValueFor('siteFooterAccent'),
+  footerBorderThickness: getValueFor('footerBorderThickness'),
+};
+
+console.log(config);
```
joetower, line 139: :laughing: We should remove any and all `console.log` statements..

PR #450, opened 2024-12-19
IYY-285: Site editors can add a caption to Image Banner | Update Image Banner component to use caption

joetower:
### Description of work
- Add image caption to image banner component
- Atomic: https://github.com/yalesites-org/atomic/pull/292
- YaleSites Platform: https://github.com/yalesites-org/yalesites-project/pull/833

### Testing Link(s)
- [ ] Navigate to the [Image Banner Component Story](https://deploy-preview-450--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--image-banner)

### Functional Review Steps
- [ ] Navigate to: https://deploy-preview-450--dev-component-library-twig.netlify.app/?path=/story/molecules-banners--image-banner
- [ ] Verify a caption renders under the image
- [ ] Verify you can change the caption using the Storybook addons

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/design/HE3hKU1Pz08cOtktqiOQIx/IYY-Design?node-id=711-5747&node-type=frame&t=oOpOGDcaL6KOwblM-0)

### Accessibility Review


codechefmarc (approved):
Looks great!

dblanken-yale (commented):
@joetower So far I've only looked at the storybook version, but noticed that the image bounds based on whether the caption is there or not.  Is that by design?  Would it make sense to keep the image in tact and just add the caption to the bottom so that the image doesn't change?  If that was requested, that's fine, too.

https://github.com/user-attachments/assets/d14fb4aa-7b03-4844-a54d-7a1d7343b557

joetower:
@dblanken-yale That is because the Image Banner re-used code from the Grand Hero. The image is meant to fill the viewport height. I wasn't sure if this was intentional, either, based on the Image Banners creation. You added the Image Banner, correct? Was there a use case to re-use the grand hero styles? As the image banner exists now, the caption needs to be absolutely positioned over the bottom of the image since the image's visible portions change depending on screen size. 

We could definitely refactor the image banner to simply let the image be an image, and not enforce a viewport-based height. This way, the caption would simply appear under the image. Do you feel the image banner is able to be refactored? I wasn't sure how heavily it is being used, do you know?

joetower:
@dblanken-yale When you have a change, could you review what I wrote above and let me know what your thoughts are?

ntay90 (approved):

dblanken-yale:
I'll defer to @miketullo95 for sure, but imo we can refactor that.  I think it'll make it easier to make changes in the future to it.

dblanken-yale (approved):

PR #458, opened 2025-01-23
YSP-758 :: Add taxonomy filter drop downs in Views block

laura-johnson:
### Description of work
- Updates css for the exposed filters so that it the fields wrap if there are more than three filters present.

### Testing Link(s)
- [ ] I added a test page here: https://pr-852-yalesites-platform.pantheonsite.io/new-vocabs-test-page

### Functional Review Steps
- [ ] Verify that if there are more than three filters added to the view block, the filters wrap.

### DIFF
--- a/components/01-atoms/forms/_yds-form.scss
+++ b/components/01-atoms/forms/_yds-form.scss
@@ -210,5 +210,15 @@
     .form-item {
       max-width: 50%;
     }
+
+    // If there are more than 4 children it should wrap.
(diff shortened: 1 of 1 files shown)


dblanken-yale (approved):
Codewise this is looking good.  I assumed this was needed for the rest of 758, so marked up the metadata so others can chime in, too.  Great work!

miketullo95 (approved):

PR #497, opened 2025-04-16
YSP-822: Add Event Time to Event Display Modes

dblanken-yale:
## [YSP-822: Add Event Time to Event Display Modes](https://yaleits.atlassian.net/browse/YSP-822)

In order to properly display the time correctly in the views, a modification to how we handle date times to accommodate more logic on whether it's the same day or not, but different time was needed.

### Description of work
- Added new formats for same-day different time scenarios.
- Fixed formatting inconsistencies by directly using `em_dash`.
- Introduced `same_date` check to handle same-date formatting cases.
- Updated logic to ensure proper formatting for start and end times.

This enhances the flexibility and correctness of date-time formatting.

### Testing Link(s)
- [ ] Navigate to the [Component Story](https://deploy-preview-497--dev-component-library-twig.netlify.app/)

### Functional Review Steps
- [ ] Verify that dates and date times work as intended in different components utilizing it

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/design/bTjNHdiy1OdgHrP3b9m7uc/Yale-UI-Kit?node-id=11249-6771)

### Accessibility Review

### DIFF
--- a/components/01-atoms/date-time/yds-date-time.twig
+++ b/components/01-atoms/date-time/yds-date-time.twig
@@ -46,15 +46,17 @@ format__day__long = Thursday, December 3, 2024
     'format__iso': 'c',
     'format__html_date': 'Y-m-d',
     'format__date': 'F j, Y',
-    'format__date__start': 'F j, Y' ~ em_dash,
+    'format__date__start': 'F j, Y',
     'format__day__full__start': "D M j \t\o",
     'format__day__full': 'D M j, Y',
-    'format__time__start': 'g:i a' ~ em_dash,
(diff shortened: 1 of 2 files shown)


dblanken-yale:
Holding off on this till we have more time to test all cases.  We need to think about this as a way for any date/time to be processed through this.  More time is needed to determine the best output for this such that it can be properly used.

For instance, in the current event meta block, we are using this, but we're using it multiple times--once for the day, and once for the the time.  And if a date spans multiple days, that doesn't output good data.

We need to know that all dates are taken care of in one place.

PR #499, opened 2025-04-16
YSP-589: Add Page Content Type Support to Reference Card Block

dblanken-yale:
### Description of work
- Updated the grid-column property for the reference card to improve layout alignment in two-column configurations, restoring the previously used alignment for single column configurations. Added a specific rule for `.yds-two-column` to ensure proper positioning/overriding.

### Testing Link(s)
- [ ] Navigate to the [Post Card Story](https://deploy-preview-499--dev-component-library-twig.netlify.app/?path=/story/molecules-cards--post-card&args=collectionType:single)

### Functional Review Steps
- [ ] Change the `Collection Type` to `single`
- [ ] Verify that the overlay to the image looks mostly right justified such that you can see over 50% of the image.

### Design Review
- [ ] Verify the designs match the [Figma Designs](https://www.figma.com/design/bTjNHdiy1OdgHrP3b9m7uc/Yale-UI-Kit?node-id=11249-6771)

### Accessibility Review

### DIFF
--- a/components/02-molecules/cards/reference-card/_yds-reference-card.scss
+++ b/components/02-molecules/cards/reference-card/_yds-reference-card.scss
@@ -363,9 +363,13 @@ $break-single-card-collection-max: tokens.$break-xl - 0.05;
     }
 
     @media (min-width: tokens.$break-xl) {
-      grid-column: 2 / -1;
+      grid-column: 3 / -1;
       grid-row: 2 / 7;
       align-self: center;
+
+      .yds-two-column & {
+        grid-column: 2 / -1;
+      }
     }
   }
 
--- a/components/02-molecules/cards/reference-card/examples/page-card.yml
+++ b/components/02-molecules/cards/reference-card/examples/page-card.yml
@@ -1,0 +1,7 @@
+reference_card__heading: Your Story, Beautifully Told
+reference_card__snippet: YaleSites is purpose-built for handling data classified as low risk, ensuring a secure environment for your web content needs.
+reference_card__url: 'https://google.com'
+reference_card__categories:
+  - content: Page
+reference_card__tags:
+  - content: Page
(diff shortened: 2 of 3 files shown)


miketullo95:
@dblanken-yale is there a specific page i can see a page reference card? instructions above say to go to the post card story. Its probably fine for now since we want to get this in for the release. Also need some do good work for this stuff: 
Screenshot 2025-04-16 at 4 49 42 PM

dblanken-yale:
Not in storybook, but in the drupal instance.  I'll see if I can add that though for completeness.  Without a reference category already established, where do you feel it should live?  Next to the post card one?

dblanken-yale:
I did just add a "Page card" in storybook showing it off.  I am going to spend a little time cleaning up that controls interface while I'm here.

dblanken-yale:
> > Not in storybook, but in the drupal instance. I'll see if I can add that though for completeness. Without a reference category already established, where do you feel it should live? Next to the post card one?
> 
> @dblanken-yale i've thought about this more, i wonder if this will just be a new display mode in views and the 'reference card' is more of a drupal component. Maybe in the future the block reference card can have the other display modes like card, list, and condensed too. I'm wondering if that was the original thought process and thats why reference card didn't have its own page

I do agree that "single" doesn't mean much with respect to the storybook instance; definitely feels "Drupal".  Maybe if we had a different name for it here?  But maybe not change it now?

miketullo95:
> > > Not in storybook, but in the drupal instance. I'll see if I can add that though for completeness. Without a reference category already established, where do you feel it should live? Next to the post card one?
> > 
> > 
> > @dblanken-yale i've thought about this more, i wonder if this will just be a new display mode in views and the 'reference card' is more of a drupal component. Maybe in the future the block reference card can have the other display modes like card, list, and condensed too. I'm wondering if that was the original thought process and thats why reference card didn't have its own page
> 
> I do agree that "single" doesn't mean much with respect to the storybook instance; definitely feels "Drupal". Maybe if we had a different name for it here? But maybe not change it now?

agree, this is a cleanup ticket for the future. maybe its 'featured'? but i know there's two different treatments within reference card that uses 'featured'

miketullo95 (approved):

PR #605, opened 2026-01-09
feat: 1012 demo for color palette update - DEMO ONLY, DO NOT MERGE

laura-johnson:
## [Y1012: OPAC Change Color Palette](https://github.com/yalesites-org/YaleSites-Internal/issues/1012)

**Warning: We did some hackery to get tokens branches to work with this repo.  Please undo dblanken-yale's commits before merging**

### Description of work
New Haven Green Slot 1 -> #4A4A4A
- The current value of that slot is color.gray.700.value. #4A4A4A is so similar to the current gray there that I thought we should just update the value for color.gray.700.value (which I did). This does affect other themes, notably:
-- Default text color across the site
-- Component theme "two" background and slot-one

Do we want to create a new gray instead?

ONHA Slot 3 -> #E7F1F7
- Create a new color called 'Onha Light' under blue: color.blue.onha-light.value

AI Slot 3 -> #FFD55A
- Change the value of ai-gold to this

AI Slot 4 -> #63AAFF (token in Old Blues)
- Use color.blue.light.value

AI Slot 5 -> #4A4A4A
- Use color.gray.700.value which is updated (see above)

### Functional testing steps:
- [ ] 

Tokens PR: https://github.com/yalesites-org/tokens/pull/93
Demo PR: https://github.com/yalesites-org/yalesites-project/pull/1184


dblanken-yale:
@laura-johnson Revert my commit when we're ready to merge.  package-lock.json was attempting to use git+ssh which it can't on CI for tokens.  using git+https it can do it.  I also added the build directory to your tokens repo, so don't forget to undo that commit also.  I'm going to modify the descriptions in each so that we now and not accidentally merge those.

PR #636, opened 2026-04-25
YISP-101: Resource Publications View Mode Enhancements

square360-george:
### Description of work
- Adds `publication-meta` molecule with shared visual treatment for resource card/list-item/condensed meta labels (journal, discipline)
- Adds `publication-detail` molecule for resource full-view mode with metadata grid, abstract, description slot, and video embed
- Converts metadata grid to definition lists for improved accessibility and semantic structure
- Adds description slot to publication-detail rendered after abstract and before video embed
- Moves video embed above metadata grid in publication-detail so video resources lead with the player
- Updates text-copy-button to support optional wrapper element (defaults to `<span>`) and restores original text on copy reset, with citation usage example in Storybook
- Fixes video embed responsive display
- Adds `--related` modifier to card-collection for related-content sections with adjusted top margin and h3 visual heading size

### Other work completed in:
- [YISP 101: Enhance the Resource Content Type for Publications](https://github.com/yalesites-org/atomic/pull/452)
- [YISP-101: Enhance the Resource Content Type for Publications](https://github.com/yalesites-org/yalesites-project/pull/1233)

### Testing Link(s)
- [ ] Navigate to the [Publication Detail Story](https://deploy-preview-636--dev-component-library-twig.netlify.app/?path=/story/molecules-meta-publication-detail--publication-detail)
- [ ] Navigate to the [Publication Meta Story](https://deploy-preview-636--dev-component-library-twig.netlify.app/?path=/story/molecules-meta-publication-meta--publication-meta)
- [ ] Navigate to the [Text Copy Button Story](https://deploy-preview-636--dev-component-library-twig.netlify.app/?path=/story/atoms-controls--text-copy-button)
- [ ] Navigate to the [Card Collection Story](https://deploy-preview-636--dev-component-library-twig.netlify.app/?path=/story/organisms-card-collection--card-collection)

### Functional Review Steps
- [ ] Verify publication-detail renders metadata grid as definition lists with label/value pairs
- [ ] Verify description slot appears after abstract and before video embed when populated
- [ ] Verify video embed displays above the metadata grid for video resources
- [ ] Verify text-copy-button copies text and restores original label after reset
- [ ] Verify card-collection `--related` modifier applies correct top margin and heading size

### Design Review
- [ ] Verify publication-meta label styling matches reference-card overline treatment
- [ ] Verify publication-detail layout matches expected full-view resource design
- [ ] Verify card-collection `--related` variant spacing is appropriate for below-content positioning

### Accessibility Review
- [ ] Verify metadata grid uses proper `<dl>`, `<dt>`, `<dd>` markup
- [ ] Verify text-copy-button announces state changes to assistive technology
- [ ] Verify card-collection `--related` heading maintains semantic h2 while displaying at h3 size
- [ ] Verify video embed is keyboard accessible and has appropriate labels


square360-george:
@dblanken I was in the Resource CT, and the copy button was bugging me. I had a look and did some tweaking.

What we did to text-copy-button on YISP-101 (on top of your PR #632):

- Atom JS — capture original button label. Replaced the hardcoded '(Copy)' revert text with a defaultLabel captured from elem.textContent.trim() at attach time. Means the button now reverts to whatever label the consumer rendered (your email-copy (copy) lowercase reverts correctly; my Copy citation reverts correctly). Comment in the JS explains why.
- Atom twig — optional text_copy_button__pre_text_tag parameter. Defaults to 'span' (backward-compatible, no consumer needs to update). Pass 'div' when pre_text contains block-level markup like rich-text paragraphs. Validates the HTML for cases where the source is not a simple string.
- Storybook — added textCopyButtonRichText story showing the pre_text_tag: 'div' use case with a multi-paragraph citation. Updated controls.mdx with a Rich-text source subsection, a new optional-props row for pre_text_tag, and an implementation-note paragraph documenting the captured-defaultLabel behavior.
- Consumer wiring (publication-detail molecule). The Resource full-page citation cell now embeds your atom (with pre_text_tag: 'div' for the rich-text citation HTML). This let me delete the bespoke publication-detail.js (which had the broken aria-live / textContent-collision pattern that the YISP-101 a11y audit flagged as A11Y-02 — the same issues you diagnosed in your e6f3f9ff commit). Also dropped the now-unused publication-detail: library entry from atomic.libraries.yml.
- Took your e6f3f9ff finding to heart — didn't add an aria-live region or aria-label.

Net: branch is a strict superset of develop. Two small, additive enhancements to your atom (defaultLabel + pre_text_tag), zero regressions, zero conflicts.

components/02-molecules/meta/publication-meta/yds-publication-detail.twig
```diff
@@ -0,0 +1,247 @@
+{#
```
dblanken-yale, line 1: We recently changed how Storybook is structured, and I'm noticing the stories for publication-detail aren't present yet. Can you add them before we merge? You'll need a data YAML, props YAML, stories file, and MDX doc — take a look at how other components like resource-meta/ or event-meta/ are set up for reference. This is what allows folks outside of Drupal to use Storybook as the source of truth for how these components work.

components/02-molecules/meta/publication-meta/yds-publication-detail.twig
```diff
@@ -0,0 +1,247 @@
...
+            <dt {{ bem('field__label', [], publication_detail__base_class) }}>
+              Abstract
+            </dt>
+            <dd {{ bem('abstract__value', [], publication_detail__base_class) }}>
+              {{ publication_detail__abstract }}
+            </dd>
+          </dl>
+        {% endif %}
+
+        {% if publication_detail__description %}
```
dblanken-yale, line 240: Every other metadata field in this template uses `<dl><dt><dd>` for semantics, but the description section uses plain `<div>` wrappers. The abstract section renders similar content and uses `<dl>`. Is there a reason description is different here? If it should have a label like "Description" (similar to how abstract has "Abstract"), should it follow the same `<dl>` pattern for consistency and screen reader support.

components/02-molecules/meta/publication-meta/_yds-publication-meta.scss
```diff
@@ -0,0 +1,104 @@
...
+  font-variant-numeric: oldstyle-nums;
+  letter-spacing: 0.5px;
+}
+
+@mixin publication-label {
+  @include publication-text;
+
+  font-weight: var(--font-weights-mallory-book);
+  font-family: var(--font-families-mallory);
+  font-size: 0.85em;
```
dblanken-yale: This is the only hardcoded font size in the publication SCSS — everything else uses tokens or CSS custom properties. Does it make sense to try to replace `0.85em` with an appropriate token value like `var(--font-size-14)` or similar?  Just trying to rely on tokens as much as possible.

components/02-molecules/meta/publication-meta/_yds-publication-detail.scss
```diff
@@ -0,0 +1,259 @@
...
+    }
+  }
+
+  /* Full-width cells */
+  &__authors,
+  &__citation,
+  &__description,
+  &__abstract {
+    @media (min-width: tokens.$break-l) {
+      flex: 1 0 100% !important;
```
dblanken-yale: This `!important` is needed to override the `> *` rule above. The event-meta component avoids this by using specific class selectors that naturally win on specificity. https://github.com/yalesites-org/component-library-twig/blob/fd03b16c7147b51e6a09b5b44ff733e34f8a50ed/components/02-molecules/meta/event-meta/_yds-event-meta.scss#L130-L148 Would something like `&__details-wrapper > &__authors` work here instead? Not a blocker, just a consistency note.

components/02-molecules/meta/publication-meta/_yds-publication-detail.scss
```diff
@@ -0,0 +1,259 @@
...
+
+    a {
+      @include atoms.plain-link;
+      @include tokens.body-s;
+
+      color: var(--color-text);
+    }
+  }
+
+  /* Abstract value and citation source text share the same typography.
```
dblanken-yale: This block is repeated 3 times for `abstract`, `description`, and `citation`. I see the comment about the stylelint specificity rule — would a private mixin like `_rich-text-cell` let you DRY this up while still satisfying the linter? Not a blocker, just something to consider.

dblanken-yale (changes requested):
I had some findings, but overall great work.  I especially wanted to thank you for the text-copy-button changes, use of dl/dt/dd for metadata fields, and the `--related` modifier for card-collections! <3

components/02-molecules/meta/publication-meta/yds-publication-detail.twig
square360-george: Thanks for the careful read, — addressed the SCSS and Twig items in 7dd3cb98. Storybook stories I'm parking until the rest is accepted so we don't churn fixtures while things are still settling; will ping when they're up.

components/02-molecules/meta/publication-meta/yds-publication-detail.twig
square360-george: Good catch — 7dd3cb98 brings description in line with abstract: same <dl><dt><dd> shape with a "Description" label.

components/02-molecules/meta/publication-meta/_yds-publication-meta.scss
square360-george: Swapped in 7dd3cb98 — publication-label now uses the body-xs mixin, with the colour / letter-spacing / oldstyle-nums numerics re-asserted so the cell still reads as one group with the value text.

components/02-molecules/meta/publication-meta/_yds-publication-detail.scss
square360-george: Reworked in 7dd3cb98. Authors and citation are now scoped under &__details-wrapper > &__authors / &__citation, so they win on specificity without !important. Abstract and description sit outside the details-wrapper, so they don't need the override at all and dropped from the rule.

components/02-molecules/meta/publication-meta/_yds-publication-detail.scss
square360-george: Pulled out into a rich-text-cell mixin in 7dd3cb98. The three rules stay separate in source order, so stylelint no-descending-specificity keeps passing.

square360-george:
@dblanken-yale 
Storybook stories landed in 18950db2. Three named variants under Molecules / Meta / Publication Detail — Policy Brief (mirrors /research/publications/isps18-04), Journal Article (with publication / issue + DOI external source), and Video Resource (oEmbed iframe). Date Format control switches between Year / Month-Year / Full per-variant.

Found a double download arrow while wiring the PolicyBrief fixture and rolled the fix into the same commit: lib/ys_link's download decorator was stamping a fa-circle-down on top of the manual circle-down the molecule was including, because the molecule's _yds-icon.twig SVG doesn't carry the .fa-icon class that ys_link's hasNoIcon() check looks for. Dropped the manual icon and let ys_link own the decoration. Worth a glance when you re-review.

dblanken-yale (approved):
Think we have something worth giving to the broader testers.  Thanks!

PR #649, opened 2026-06-18
1161: Views Block Rework

dblanken-yale:
### Description of work
Links use an underline-skip `text-shadow` whose color must equal the background so the animated underline weaves behind the letters. The shadow defaults to white (overridden only for the dark footer / single-card collections) and was never matched to `[data-component-theme]` sections — so in a themed section (e.g. navy) the white shadow showed as a halo around the link text.

- **text-link** (`_yds-text-link.scss`): override `--color-text-shadow` to `var(--color-background)` under `[data-component-theme]` (covers heading links / CTAs).
- **reference-card** (`_yds-reference-card.scss`): same override for the card heading link (`.reference-card__heading-link`), so listing **card titles** in a themed section no longer halo. Mirrors the existing single-collection override.
- Surfaced while QA'ing the Views Block rework; affects all links/card titles in component-themed sections, not just listing blocks. Plain/white pages and the dark footer are unaffected.
- Other work completed in: yalesites-org/yalesites-project#1299

### Round-3 QA fixes (from the third review round on PR #1299)
- **Filter bar uniform widths + Apply row** (`_yds-form.scss`): the scaffold exposed-filter controls rendered at ragged widths in constrained sections. Replaced the content-sized widths and ad-hoc `:has()` min-width rules with a uniform column grid (`repeat(auto-fit, minmax(min(100%, 13rem), 1fr))`) sized to the form's own inline width — 2 equal columns in a 50/50 section, more on wider sections, 1 when narrow — and moved the Apply button to its own full-width row.
- **AA contrast on themed sections** (`_yds-pager.scss`, `_list-mixins.scss`, `_yds-list.scss`): list-view result separators, the pager, and the category/tag term labels used fixed colors that failed WCAG AA on themed section backgrounds (slots 2/4/5). Routed each through the section's on-theme tokens (`--color-layout-content` / `--color-layout-theme` / `--color-layout-border`) with fallbacks that leave the default appearance unchanged. The pager's current-page marker previously used the muted `--color-basic-brown-gray` (~4.87:1 on the default background); it now uses the theme foreground (~16:1 default, AA on every theme) with the solid border still marking the current page.
- Verified WCAG AA across themes in Storybook: pager 15.03 / 16.10 / 5.34, separators 11.27 / 15.03 / 4.98, term labels 15.03 / 16.10 / 5.34; default (non-themed) appearance unchanged.

### Functional testing steps:
- [ ] Place a listing in a section with a non-white component theme (e.g. navy/"one"); confirm card titles no longer have a white halo.
- [ ] Add a heading link to such a block; confirm it has no halo either.
- [ ] Confirm links/card titles on plain (white) pages are unchanged.
- [ ] Confirm dark-footer links are unchanged.
- [ ] Place a listing with 3+ exposed filters in a 50/50 (and 70/30) section; confirm the filter controls are all the same width, wrap cleanly to two per row, and the Apply button sits on its own full-width row.
- [ ] In a themed section (slots 2/4/5), confirm the list-view result separators, pager (links, hover, current page), and category/tag term labels all meet contrast and stay legible.

References yalesites-org/YaleSites-Internal#1161

Created with AI


dblanken-yale:
## Chip label clipping fix (QA round on yalesites-org/yalesites-project#1299)

`7104e15b` — selected-term chip labels were losing their descenders ("Faculty" lost the base of its "y") in the listing filter bar.

The reported cause was the chip rendering taller than the `3.125rem` control could hold. It is not: the chip is 24px inside a 50px control and nothing overflows the container (`.chosen-choices` has `scrollHeight === clientHeight`). The clipping is inside the chip — Chosen's vendor CSS pins `line-height: 12px` on `li.search-choice` while its `font-size: 14px` on `.chosen-container` is overridden in our scaffold context, so the label lands at 95% of the form's ~20px: 19px of text in a 12px line box. The `span { overflow: hidden }` we added for tail-ellipsis truncation then clips the overflowing glyph ink (measured `clientHeight 12` vs `scrollHeight 18`). Being ink overflow rather than box overflow is why box geometry reported nothing wrong and only labels with descenders showed it.

Fixed with our existing `--font-style-body-xs` token on the chip. It stays a `font` shorthand on purpose: the line-height gives the label a line box that fits its ink, and the smaller scale keeps the chip inside the 3.125rem control height — raising the control would break the filter/search alignment, and a line-height alone leaves the chip too tall. Verified in a browser: `clientHeight === scrollHeight`, chip 33.2px, control still 50px, single-line `+N more` state unchanged.

The `+N more` badge got the same token so it does not out-size the chips it stands in for (medium weight preserved by declaring the shorthand before `font-weight`). That part is a visual call rather than a bug fix — one line to drop if @miketullo95 prefers the badge larger.

Gates: `stylelint`, `eslint`, `prettier`, `build` all exit 0.

Other work completed in:
- yalesites-org/yalesites-project#1299 — config modal re-measure fix (`b24121bd`)

dblanken-yale:
## Selected-term chip no longer breaks its container's bounds (QA round on yalesites-org/yalesites-project#1299)

`c5d101b0` — a selected term in a listing filter painted over the bottom border of its own box, so the control read as broken open underneath the pill.

@miketullo95's report put it on the pill ("the pills look too tall"). Measured, the pill is fine — 33.2px in a 50px control, `clientHeight === scrollHeight`, so the `--font-style-body-xs` token from `7104e15b` is doing its job. The fault is the container:

- `.chosen-container` is 50px border-box with 1px borders **and 10px block padding**, leaving a **28px** content area.
- That padding is not ours by intent. Chosen copies the source `<select>`'s classes onto the container it generates (`inherit_select_classes`), so `.form-item__select`'s padding — sized for a native select's single line of text — lands on the Chosen box. Same class-copying cause as the `min-width: 0` reset already in this scope.
- The chip row is **39.2px** (33.2px chip + Chosen's 3px block margins), so it ran **1.2px past the container's inner bottom edge**, and being opaque white it painted over the 1px bottom border.

Fix: `padding-block: 0` on `.chosen-container` in the scaffold scope, plus `min-height: 100%` on `.chosen-choices` so its existing `align-items: center` centres the chip across the whole control rather than in a shrink-to-fit row at the top. `min-height` and not `height` on purpose — a percentage resolves only against the container's definite closed height, so it is inert in the open state where `.chosen-with-drop` sets `height: auto` and wrapped chips still grow the box. `min-height` on the container is the matching floor: without it, releasing `height` in the open state made an opened filter *shorter* than its closed self.

It cannot be folded into the neighbouring `.form-item__select` min-width reset, and that is worth knowing before someone simplifies it: `min-width: 0` is harmless on a native `<select>`, but zeroing block padding is not — and `chosen.settings` only initialises a multi-select at 20+ options, so shorter filters really do render natively here.

Measured after, logged out: container 50px, **overflow past the border 0.00px** (was +1.2px), chip 33.2px centred at 7.39px / 7.41px. Four chips: closed 50px with the "+N more" badge, open 108.4px with every chip wrapped and no overflow, re-closed back to 50px.

The repeated `3.125rem` became `$filter-control-height`. Deliberately **not** applied to the filter bar's search field — that is styled by the global `.form-item-search .form-text` rule, which is not scaffold-scoped and would drag every other search box on the platform with it; the comment says so, and says the two values have to be changed together. Verified behaviour-neutral: comparing the compiled bundle before and after the extraction, **0 of 3400 selectors changed their declaration set**.

Gates: `lint` (eslint + stylelint), `prettier`, `build` all exit 0.

Note for testing: this only reaches an environment after that environment rebuilds `dist/` from the matching `1318-views-rework` branch. My local `dist` was stale and showed the pre-fix chip until rebuilt.

**Accessibility, flagged not fixed:** Chosen's per-chip remove control (`.search-choice-close`) has an empty accessible name (confirmed from the browser's accessibility tree) — WCAG 2.1 AA 4.1.2, one per selected term on public pages. The natural fix is here, in `yds-select.js`'s `updateChoiceSummary()`, which already walks every chip and already has the term label. I left it out of this round because it is pre-existing, unrelated to this fix, and that function is the chip-summary logic that has taken several QA rounds. Small standalone follow-up.

Other work completed in:
- yalesites-org/yalesites-project#1299 — front-end dark-mode leak fix (`6cecf7ecc`)

miketullo95:
**TL;DR:** Tagging this `work in progress` and pulling `needs review`. This is the epic branch for the Views Block Rework, not a review unit yet.

Sub-task PRs are still merging into this branch and building the epic up. Reviewing it now would mean reviewing a moving target, so we're holding off until the last child lands.

When the epic is complete we'll swap `work in progress` back to `needs review` and do a full pass on the whole body of work. No action needed here in the meantime.

PR #668, opened 2026-07-20
998: Optimize Gallery Block Image Loading and File Size Limits

dblanken-yale:
### Description of work
- Defer the interactive gallery lightbox images until the modal is opened. `.media-grid__modal` is a full-viewport `position: fixed` overlay that was hidden with `visibility: hidden` while inactive; because a `visibility: hidden` element still occupies layout, the browser treated the modal's `loading="lazy"` images as on screen and fetched every sized derivative on page load. This defeated the ticket goal even after the companion yalesites-project PR repointed the modal at a sized (`image_max_width`) derivative — all of them still downloaded up front.
- Hide the inactive modal with `display: none` instead, so native lazy loading defers the modal images until the lightbox is opened. The open/close fade is preserved with a `display ... allow-discrete` transition plus a `@starting-style` entry opacity.
- Because the closed modal is now removed from layout, the caption-truncation script could no longer measure a long caption's collapsed height at attach time (`offsetHeight` is 0 under `display: none`, which would clip long captions to zero height on open). The collapsed-height measurement is deferred to the first time the modal opens.
- **Single work unit** with yalesites-org/yalesites-project#1374 (gallery media view modes + config) and yalesites-org/atomic#490 (thumbnail view mode). All three share this branch name and must be reviewed and merged together — this change is what makes the "don't eager-load modal images" goal actually hold at runtime.
- Other work completed in: yalesites-org/yalesites-project#1374, yalesites-org/atomic#490

### Functional testing steps:
- [ ] On a page with a Gallery block, load the page fresh and do **not** open a modal. In DevTools Network, the gallery modal's `max_width_*` derivatives are **not** requested on page load (equivalently, every `.media-grid-modal__media img` has `naturalWidth: 0`).
- [ ] Open a gallery image in the lightbox — the modal image now fetches its `max_width_*` derivative **on open**, and the modal fades in.
- [ ] Close the modal — it fades out and returns to hidden; reopening still works, and navigating between items (prev/next, pager, swipe) works.
- [ ] On a gallery item with a long caption (over 100 characters), open the modal — the caption heading/area renders at full height (not clipped to zero), and the expand ("+") toggle reveals the full caption text.
- [ ] With `prefers-reduced-motion: reduce`, the modal still opens and closes correctly (no fade animation).

References yalesites-org/YaleSites-Internal#998

### DIFF
--- a/components/03-organisms/galleries/media-grid/_yds-media-grid-modal.scss
+++ b/components/03-organisms/galleries/media-grid/_yds-media-grid-modal.scss
@@ -25,29 +25,49 @@ $modal-speed: var(--animation-speed-slow);
   width: 100%;
   background: transparent;
 
-  [data-media-grid-modal-state='inactive'] & {
-    @include tokens.animate-hidden($modal-speed);
-    @include media-grid-modal-transition;
+  // Fade the modal in/out. `display` is transitioned with `allow-discrete` so
+  // the inactive modal can use `display: none` (see below) while still animating
+  // opacity on open and close.
+  @media (prefers-reduced-motion: no-preference) {
+    transition: opacity $modal-speed ease-in-out,
+      display $modal-speed ease-in-out allow-discrete;
+  }
 
+  [data-media-grid-modal-state='inactive'] & {
+    // Use `display: none` (not `visibility: hidden`) so the browser removes the
+    // modal from layout entirely. The modal is a full-viewport fixed overlay, so
+    // when it merely occupies layout the browser treats its `loading="lazy"`
+    // images as on screen and fetches them on page load even though the modal is
+    // closed. `display: none` lets native lazy loading defer those images until
+    // the modal is opened, and also removes the modal from the a11y tree and tab
+    // order while inactive.
+    display: none;
     opacity: 0;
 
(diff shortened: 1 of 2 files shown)


miketullo95 (approved):
Approving. This is the fix that actually resolves the blocker I flagged on yalesites-project#1374 last round. Verified live on the pr-1374 multidev:

- Fresh load, no modal opened: 0 `max_width_*` requests, all modal images `naturalWidth: 0`. `.media-grid__modal` computes to `display: none` while inactive.
- Opening the lightbox fetches the image on open and renders it, modal fades in.
- Prev/next and close work correctly, close returns the modal to `display: none`.

Also read through `_yds-media-grid-modal.scss` and `yds-media-grid-interactive.js`: the `display ... allow-discrete` + `@starting-style` fade and the caption-height-on-first-open logic (`truncatedCaptions` deferred measurement in `toggleModalState`) both look correct. Wasn't able to exercise the long-caption case live since this multidev's seeded gallery content doesn't have an item with a caption over 100 characters, but the code path checks out.

Good to merge alongside yalesites-project#1374 and atomic#490.

@dblanken-yale

PR #669, opened 2026-07-20
779: New Block: Audio Component

dblanken-yale:
### Description of work
- Fix the audio player review defects flagged on component-library-twig#468 so the component can be merged:
  - Extract a shared `range-slider` mixin so the progress and volume sliders can no longer drift apart; remove the non-standard `appearance: none` from the webkit progress track (the likely cause of the reviewer's "can't see the line" symptom in real Safari); unify the thumb sizes across engines; give the range input an explicit height so the thumb is never clipped; and center the thumb with a computed margin.
  - Hide the volume control below the mobile breakpoint (volume does not function on mobile; physical device volume handles it) — the locked decision on the issue.
  - Make the whole speed control area a click target so the visible active label (e.g. "1x") opens the options, not just the gauge icon.
- Convert the Storybook story to the current convention: a new `audio-props.yml` consumed via `_storybook/component-props`, with the shared tags — matching the other atom stories (e.g. video-embed).
- **Review rework (2026-07-21):**
  - Fix the `audio_embed__alighment` → `audio_embed__alignment` typo in `yds-audio.twig` so Post/Event alignment works (pairs with yalesites-org/atomic#491, which passes the value through).
  - Make the player theme-aware: swap the fixed `--color-gray-*` / `--color-blue-medium` tokens for the ambient theme tokens (`--color-background` / `--color-text` / `--color-link-base`) so it is legible on dark pages instead of a light box; ring the range thumb in the page background so the handle stays distinct from the track. Scoped to the platform light/dark theme tokens, not per-section custom backgrounds.
  - Rename the story title `Atoms/Audio` → `Atoms/Audio Player` (and the MDX Overview heading) for naming consistency.
  - Reset the player to the paused state when playback ends (listen for the audio `ended` event) so the play icon returns instead of the pause icon lingering at `m:ss/m:ss` — the clip can then be replayed.
- Other work completed in: yalesites-org/atomic#491
- Other work completed in: yalesites-org/yalesites-project#1396

### Verification
- Slider rendering verified across Chromium, Firefox, and WebKit via Playwright (before/after screenshots).
- Speed control verified with a Playwright interaction test: clicking the visible "1x" label opens the options, selecting a speed closes the menu, the gauge still toggles, and `aria-expanded` tracks state.
- All controls confirmed keyboard-focusable with accessible names (play/pause, progress, speed toggle, volume).
- Theme-awareness verified against the built `dist` CSS in light (`:root`) and dark (`[data-basic-theme]`) contexts: on dark the box background and text flip so the label/controls stay legible (light-box bug gone) and the thumb keeps a visible ring on the track. Chromium only locally.

### Caveats and open questions
- Cross-browser hardening still needs a **manual QA pass on real Safari, Firefox, Edge, iOS Safari, and Chrome Android** — the original "invisible slider track" symptom (real Safari, 2025) did not reproduce locally on Chromium/Firefox/WebKit, and the thumb-ring border should be eyeballed cross-engine (native range-thumb box-sizing varies).
- Theme-awareness note for @miketullo95: to be theme-aware without introducing `color-mix` (not used elsewhere in the library), the box now uses the page background with a 1px themed border instead of the old fixed light-gray fill — so on light pages the box reads as bordered rather than filled. If you prefer to keep the subtle filled surface in both themes, a `color-mix(...)` tint is the alternative — happy to switch.

### Functional testing steps:
- [ ] In Storybook, open **Atoms/Audio Player** and confirm the progress and volume tracks + thumbs are visible and draggable.
- [ ] Click the visible "1x" speed label (not just the gauge icon) and confirm the 0.5x / 1x / 2x options open; pick a speed and confirm the menu closes.
- [ ] Resize to mobile and confirm the volume control is hidden while the progress bar remains usable.
- [ ] View the player on a **light** page and a **dark-themed** section: on dark the box background and text flip so the label and controls stay legible (no light box on dark), and the progress/volume thumbs keep a visible ring against the track (thumb not merged into the track).
- [ ] Let a clip play to the end (or seek to the end) and confirm the control returns to the **play** icon (not a lingering pause icon), and that pressing play replays from the start.
- [ ] Cross-browser QA the sliders on real Safari, Firefox, Edge, iOS Safari, and Chrome Android.
- [ ] Confirm **Audio Player** now appears under **Atoms** in the Storybook sidebar and opens an Overview doc page like the other atoms.
- [ ] On the Audio Player Overview page, confirm the interactive preview renders the player and the Properties tables list the required Audio URL and optional Content props.

References yalesites-org/YaleSites-Internal#779

---
Created with AI


dblanken-yale:
I am not seeing the audio story in the atoms in the storybook for this PR.  We will ensure that it displays properly.  I also think we're missing a mdx file for it like the others have.  We'll make sure it adheres to our standards going forward.

https://deploy-preview-669--dev-component-library-twig.netlify.app/

dblanken-yale:
## Reworked the review feedback

@dblanken-yale — thanks for catching this. Both points had the same root cause and are fixed in `a63ea39b`.

- **Audio story not appearing under Atoms + missing MDX file** (@dblanken-yale) — Every atom story sets `tags: ['!dev']` (which hides the auto-generated stories from the sidebar) and relies on an `.mdx` `<Meta name="Overview">` page to register the visible sidebar entry (`docs.autodocs` is disabled globally). Audio was the **only** atom without an MDX file, so it had no `dev`-tagged entry and nothing showed under Atoms. Added `components/01-atoms/audio/audio.mdx` mirroring the sibling convention (`divider` / `video-embed`): an Overview page with the interactive preview, the shared `twigPropsTable` properties tables sourced from `audio-props.yml`, and player + accessibility documentation. This both makes **Atoms/Audio** appear in the sidebar and supplies the missing component docs to match the other atoms.

Verified locally: a clean Storybook build now emits `docs | atoms-audio--overview` (tagged `dev`, matching the other atoms), and on `npm run develop` the **Audio** entry shows under **Atoms** with the Overview page rendering the player and the Required/Optional property tables (0 console errors). The deploy preview above rebuilds from `a63ea39b`.

This change is CLT-only (Storybook); the companion PRs in the work-unit are unchanged:
- yalesites-org/atomic#491
- yalesites-org/yalesites-project#1396

miketullo95 (changes requested):
Reviewed against #779.

**Blocking**

1. `components/01-atoms/audio/yds-audio.twig:16` reads `audio_embed__alighment` (misspelled). That's the root cause of the center-alignment bug you flagged for Post/Event pages. Fix the variable name to `audio_embed__alignment`. Note this needs to land together with atomic#491, the value isn't even being passed through from the block template right now, so both fixes are needed for alignment to actually work.

2. `components/01-atoms/audio/yds-audio.scss:89` (and the `--color-gray-*` / `--color-blue-medium` tokens used for the track, thumb, and border throughout this file) are global rather than theme-aware, which is what's causing the light-box-on-dark-page issue you flagged. Scope for now: get it working against the platform's light/dark theme tokens, that's the priority and an easy win. Don't try to extend it to fully custom per-section background colors in this PR, that's a bigger lift with its own testing burden, worth a follow-up if it comes up again.

**Non-blocking**

3. `components/01-atoms/audio/audio.stories.js:12` — title is `Atoms/Audio`. Rename to `Atoms/Audio Player` to match the "Audio Player" naming everywhere else.

@dblanken-yale

dblanken-yale:
## Reworked the review feedback

- **`audio_embed__alighment` typo** (@miketullo95) — fixed to `audio_embed__alignment` in `components/01-atoms/audio/yds-audio.twig`. Pairs with yalesites-org/atomic#491, which now passes the value through; together they make Post/Event center-alignment work. `4970ba5f`
- **Global (non-theme-aware) colors → light box on dark pages** (@miketullo95) — swapped the fixed `--color-gray-*` / `--color-blue-medium` tokens in `yds-audio.scss` for the ambient theme tokens: the box uses `--color-background` + `--color-text` with a 1px `--color-text` border; the slider track uses `--color-text`; the thumb and accents use `--color-link-base`. Added a `--color-background` ring on the range thumb so the handle stays distinct from the track when a region remaps thumb and track to the same color (e.g. dark sections). Scoped to the platform light/dark theme tokens only, not per-section custom backgrounds, per your note. `4970ba5f`
- **Story title `Atoms/Audio` → `Atoms/Audio Player`** (@miketullo95, non-blocking) — renamed in `audio.stories.js`, and updated the `audio.mdx` Overview heading to match. `4970ba5f`

Verified light and dark rendering against the built `dist` CSS (before/after screenshots): on light the box is white with a gray border, gray track, blue thumb, blue active speed label; on dark the box background and text flip so the label stays legible (the light-box bug is gone), and the thumb keeps a visible ring on the track. Chromium only locally — the cross-browser QA on real Safari/Firefox/Edge/iOS/Android in the testing steps still applies.

Other work completed in:
- yalesites-org/atomic#491 — pass `audio_embed__alignment` into the embed (paired with the typo fix above)
- yalesites-org/yalesites-project#1396 — flat block form, Audio Player naming, 20 MB upload cap

dblanken-yale:
## Design note — light-page player background

Tracking the visual change from the theme-awareness fix so it's not a surprise in review. To make the player theme-aware without introducing `color-mix` (not used elsewhere in the library), the box now uses the page background (`--color-background`) with a 1px themed border (`--color-text`), instead of the old fixed light-gray fill (`--color-gray-100`).

Effect:
- **Dark pages:** fixed — was a light box with unreadable text; now flips to a dark box with legible text.
- **Light pages:** the box now reads as a bordered container rather than a filled light-gray surface.

If you'd rather keep the subtle *filled* surface in both themes, the alternative is a `color-mix(in srgb, var(--color-text) ~8%, var(--color-background))` tint — one line, happy to switch. @miketullo95

dblanken-yale:
## Fix — player resets at the end of playback

Reported by @dblanken-yale: when the audio reached the end, the player kept the **pause** icon showing (e.g. at `0:54/0:54`), so it looked like it was still playing.

Cause: the player never listened for the audio `ended` event, so `is-playing` was never reset — and the pause icon is the one shown while `is-playing="true"`.

Fix (`components/01-atoms/audio/yds-audio-player.js`): added an `ended` listener that resets the player to the paused state (the play icon returns) and moves keyboard focus off the now-hidden pause button. The next play restarts from the beginning, so the clip can be replayed. `283ed2c7`

Verified with a Playwright harness driving the real behaviour on the compiled component: **before** the fix, firing `ended` left `is-playing="true"` with the pause icon visible; **after**, `is-playing="false"` with the play icon visible and the pause icon hidden.

miketullo95 (approved):
Confirmed — the `audio_embed__alighment` typo is fixed, the player uses theme-aware tokens now (checked light and dark on the multidev, looks good as-is, no need for the color-mix alternative), and the Storybook title is renamed to "Atoms/Audio Player." Nice catch adding the paused-state reset on end too. Approving.

@dblanken-yale

PR #670, opened 2026-07-24
697: Add Icon Options to In-Line Message Block

dblanken-yale:
### Description of work
- Add an `inline_message__icon_name` variable to the In-Line Message component. A chosen icon from the shared icon set renders on the block; an empty value or `_none` renders no icon (a text-only message). The type-derived default (`circle-info`/`circle-exclamation`) applies only to the bare component when no icon value is supplied.
- Document and demo the option in Storybook (props, stories, MDX), sourcing the Icon control list from the shared `facts-and-figures` icon manifest so the two components stay in sync.
- Other work completed in: yalesites-org/atomic#493
- Other work completed in: yalesites-org/yalesites-project#1407

### Functional testing steps:
- [ ] In Storybook, open Molecules → Inline Message and use the Icon control to preview icon variants; confirm the chosen icon renders beside the message.
- [ ] Set the Icon control to `_none`; confirm no icon renders (text-only).
- [ ] Confirm the icon is decorative (aria-hidden) and does not convey information missing from the text.

References yalesites-org/YaleSites-Internal#697

---
Created with AI

### Review round 2 — register the two historical In-Line Message icons (2026-07-27)

`circle-info` and `circle-exclamation` added to `facts-and-figures-icons.yml`. They were already real SVGs in `images/icons/` but unregistered, so they could not be picked once the Icon field started sourcing this manifest.

### Additional functional testing steps:
- [ ] In Storybook, open **Molecules → Inline Message** and confirm the **Icon** control now offers **"Information Circle"** and **"Exclamation Circle"**; select each and confirm the matching icon renders.
- [ ] Confirm **Molecules → Facts and Figures**' icon control also lists the two new entries and renders them (they share this manifest — accepted on yalesites-org/yalesites-project#1407).
- [ ] Confirm the `_none` option still renders no icon.

Full write-up on yalesites-org/yalesites-project#1407.


dblanken-yale:
@miketullo95 I think this is well done in storybook, however, did we want to modify the visreg for inline message to have a different set of icons mixed in (but static) so that we can test this functionality in future instances as well?  I think it could go either way since it should be the same as the original icon, but if so let me know and I'll get it going.

dblanken-yale:
On the visreg question — leaving the call to @miketullo95 and @dblanken-yale, but here is one data point that may settle it.

The Percy workflow in this repo (`.github/workflows/visual-regression.yml`) only triggers on `pull_request: types: [ready_for_review]`, and `gh run list --workflow visual-regression.yml` currently returns **zero runs**. Our PRs are opened non-draft, so that trigger never fires. A new set of static icon variants in `inline-message.visreg.stories.js` therefore would not be exercised by CI as things stand today.

So: if the goal is future-proofing for whenever visual regression is switched back on, adding them is cheap and harmless. If the goal is coverage right now, it would not buy anything yet.

I have not touched the visreg story in this PR — happy to add it if you want it. This branch was also merged up with `develop` so it stays cleanly mergeable; no component changes in that merge.

The related fix for the review feedback on the Icon field lives in yalesites-org/atomic#493, with the write-up on yalesites-org/yalesites-project#1407.

dblanken-yale:
## Registered the two icons In-Line Message has always used

`circle-info` and `circle-exclamation` are now in `facts-and-figures-icons.yml`. Both have been real SVGs in `images/icons/` all along — they were simply never listed in the manifest, so once yalesites-org/yalesites-project#1407 started sourcing the Icon picker from it, the two icons the component historically rendered could not be selected. Including the one the block now defaults to.

No other change was needed in this repo: `FactsAndFiguresIconManager` reads this file directly on the Drupal side, and `inline-message.stories.js` already builds its Icon control from `Object.keys(iconsConfig.icons)`, so the Storybook control picks them up too.

They also become selectable in **Facts and Figures**, which shares this manifest. That was raised and accepted on yalesites-org/yalesites-project#1407 — easy to drop the two entries if it turns out to be unwanted.

Labels follow the file's existing sort-by-label convention: "Exclamation Circle" between Envelope and Flask, "Information Circle" between Industry Windows and Leaf. Glyph-descriptive rather than tone-descriptive, matching neighbours like "Frowning Face" and "Cloud with Rain", since these now appear in a non-message context too.

`npm run lint` exit 0; prettier clean on the changed file.

Full write-up, including the deploy-order and revision handling on the Drupal side, is on yalesites-org/yalesites-project#1407.

Other work completed in:
- yalesites-org/yalesites-project#1407 — field default + backfill deploy hook.

References yalesites-org/YaleSites-Internal#697

miketullo95 (approved):
Icon manifest addition confirmed — `circle-info` and `circle-exclamation` render correctly through the shared picker and in Storybook, matching the fix verified on the yalesites-project multidev. Approving.

@dblanken-yale

PR #671, opened 2026-07-24
#750: Add Free-Form Location Field to Drupal Events

nJim:
### Description of work
- Adds a WYSIWYG "Additional Information" fallback to the event meta/Localist location display so non-Localist events can show a free-form location.
- When a Localist place is present it always takes precedence and the free-form field is suppressed, even if both are populated
- Adds schema-org `Place`/`PostalAddress` microdata to the existing Localist location markup, and `itemprop="location"` to the free-form fallback
- Adds a new Storybook variant  so the fallback can be reviewed in isolation

### Testing Link(s)
- [ ] Navigate to the [Event Meta Component Story](https://deploy-preview-671--dev-component-library-twig.netlify.app/?path=/story/molecules-meta--event)

### Functional Review Steps
- [ ] Navigate the the [event metadata organism](https://deploy-preview-671--dev-component-library-twig.netlify.app/?path=/docs/molecules-meta-event-meta--docs)
- [ ] Check new display logic
  - [ ] In the Controls panel, set "Event Date Scenario" to `No Localist Place (Free-form Location)`
  - [ ] Confirm the Location section shows the free-form address/building/virtual-link text
  - [ ] Verify there is no map or "Get Directions" button
- [ ] Localist regression check
  - [ ] Switch "Event Date Scenario" to `Mixed (Past & Upcoming)`
  - [ ] Confirm the original Localist place, map, and "Get Directions" button still render as before
- [ ] Open browser dev tools (Elements/Inspector) on the rendered Location section and confirm:
  - [ ] When a Localist place is shown: the wrapper has `itemscope itemtype="https://schema.org/Place"` and the address has `itemtype="https://schema.org/PostalAddress"`
  - [ ] When the free-form fallback is shown, the wrapper has `itemprop="location"` and no `Place`/`PostalAddress` attributes

### Design Review
- No significant design change as elements visibility only effected. Existing treatments are employed.

### Accessibility Review
- [ ] Using a screen reader confirm the "Location" label is announced immediately before the value in both the Localist and free-form scenarios
- [ ] Confirm the added `schema.org` microdata are not announced

### DIFF
--- a/components/02-molecules/meta/event-meta/event-localist--no-place.yml
+++ b/components/02-molecules/meta/event-meta/event-localist--no-place.yml
@@ -1,0 +1,42 @@
+event_dates:
+  - formatted_start_date: Friday, May 3rd, 2024
+    formatted_end_date: Friday, May 3rd, 2024
+    formatted_start_time: 8:30 am EDT
+    formatted_end_time: 9:30 am EDT
+    original_start: 1714739400
+    original_end: 1714743000
+    is_past_event: true
+    is_all_day: false
+  - formatted_start_date: Sunday, May 10th, 2030
+    formatted_end_date: Sunday, May 10th, 2030
+    formatted_start_time: 8:30 am EDT
+    formatted_end_time: 9:30 am EDT
+    original_start: 1904483400
+    original_end: 1904487000
+    is_past_event: false
+ticket_cost: $40 pre registration, $60 at the door
+ticket_url: https://www.yale.edu
+event_meta__cta_primary__content: Event Name Website
+event_meta__cta_primary__href: https://www.example.com
+event_types:
+  - name: Music
+  - name: Lecture
+location_additional_info: '<p>Meet in the lobby of the Yale School of Music, 470 College St. This session will also be streamed live on our virtual platform for remote attendees; a link will be emailed after registration.</p>'
+event_audience:
+  - name: Audience Name one
+  - name: Audience Name two
+event_topics:
+  - name: Topic One
+  - name: Topic Two
+description: '<p>Vel quasi dicta ut fugit molestiae ea sint voluptatem soluta repellendus quod qui fugit dolores illo.</p>'
+stream_url: https://www.example.com
+stream_embed_code: Watch the event
+event_featured_date:
+  formatted_start_date: Sunday, May 10th, 2030
+  formatted_end_date: Sunday, May 10th, 2030
+  formatted_start_time: 8:30 am EDT
+  formatted_end_time: 9:30 am EDT
+  original_start: 1904483400
+  original_end: 1904487000
+  is_past_event: false
+event_featured_index: 1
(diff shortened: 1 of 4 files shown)


dblanken-yale (approved):
**TL;DR:** approving. @nJim The free-form location works and the Localist path is unchanged. Everything I found is deferred to yalesites-org/YaleSites-Internal#1767, so nothing here is blocking.

One merge constraint: do not merge this into `after-2260` ahead of yalesites-org/yalesites-project#1409, which is still `needs work`. The other order is safe. This one is not, because the template would land without the Drupal code that sets `is_localist_event`.

<details>
<summary><b>What I tested</b></summary>

I drove the deploy preview rather than only reading the diff.

- Free-form variant renders Location plus a separate "Additional Address Information" row, with no map and no Get Directions. Correct.
- Localist variant is unchanged: place, address, Show Map, expanded map, Get Directions all still work.
- axe on the free-form variant found no new violations.
- I traced the cross-repo contract into `MetaFieldsManager::getEventData()` and `ys_layouts_theme()`. The flag and both new variables are supplied correctly, and the structures match what the template reads.
- I grepped every consumer in this repo. All four fixtures were updated and both story files pass the flag through, so nothing in-repo regresses.

One note on the review steps in the description: they name a variant called "No Localist Place (Free-form Location)", which does not exist. The one you added is "Native Event (Address + Additional Info)". The Storybook Controls panel is also empty on the preview, so I had to drive the preview interface directly to switch variants. Worth correcting in the body so the next reviewer is not stuck.

</details>

<details>
<summary><b>Deferred to #1767, not blocking</b></summary>

1. The new `schema.org/Event` scope at `yds-event-meta-localist.twig:145` has no `name` and no `startDate`. I enumerated the rendered properties and both are absent. Google requires them, so Search Console will flag every event page once this ships. This is the one I would most want fixed soon.
2. The wrapper guard at line 357 ORs all four location variables regardless of the flag, but each branch reads only one pair. `.event-meta__details-wrapper` has `border-top` in `_yds-event-meta.scss:79`, so an empty row draws a stray divider.
3. Line 464 uses a bare `PostalAddress` as `location`. Valid schema.org, but Google expects a `Place` with a nested address, so this branch will not be rich-result eligible while the Localist branch is.
4. The two branches break the address differently. Localist renders one line, free-form renders two.
5. No default for `is_localist_event`. Safe today given the staging-branch setup, but a one-line `??` would make the ordering constraint above unnecessary.
6. Smaller items: the room folded into the Place `name` at line 428, an empty `itemprop="name"` span when the place has no name, the unguarded `eventDataVariants` lookup at `meta.stories.js:66`, and the stale variant count at `event-meta.mdx:22`.

</details>

<details>
<summary><b>Things I liked</b></summary>

The `location_display` fix is a real improvement. A room with no place no longer renders a leading comma. Passing `address_additional_info` as `processed_text` on the Drupal side is also the right call, since it keeps filter cacheability bubbling and avoids a `|raw` that would skip text-format filtering.

</details>

dblanken-yale:
**Heads up:** this PR changed after your approval. `dac8a72f` is the template half of the [#1409 rework](https://github.com/yalesites-org/yalesites-project/pull/1409):

- Additional Address Information now renders on Localist events too. It is labelled "Location" when nothing else rendered.
- A new `_yds-event-meta--locality.twig` partial prints city, region and postcode only when present, so non-US addresses have no stray comma.

The #1767 deferrals from your review are untouched. The merge-order constraint still stands: land yalesites-org/yalesites-project#1409 first.

PR #672, opened 2026-07-24
1435: Update Storybook to latest version and move to Vite

nJim:
### Description of work
Migrates the Storybook build from Webpack 5/Storybook 8 to Vite/Storybook 10 via Emulsify Core 4
- Bump Node to 24.13.0
- Remove Webpack config, loaders, and plugins entirely
- Add project.emulsify.json mapping the existing atoms/molecules/organisms/page-layouts/page-examples/tokens namespaces so all existing @-namespaced Twig includes keep resolving unchanged
- Move fonts/images into a root assets/ directory per the new static-asset convention
- Relocate custom Storybook setup (Twig helpers, Drupal behavior/jQuery bootstrap, YaleSites theme branding) into config/emulsify-core/
- Drop @storybook/addon-essentials and addon-styling-webpack in favor of Emulsify Core's Vite-native defaults (a11y, links, themes)
- Fix a webpack-only Sass `~` node_modules alias (18 files) and a webpack require.context() call with no Vite equivalent
- Patch the unmaintained `linkpurpose` package to add a default export since Vite doesn't support the old exports-loader shim
- Rename CommonJS config files to .cjs (required once package.json declares "type": "module")
- Remove Percy/visreg:ci wiring (no longer in use)

### Testing Link(s)
- [ ] [PR Storybook build](https://deploy-preview-672--dev-component-library-twig.netlify.app/)
- [ ] [Current production Storybook](https://yalesites-org.github.io/component-library-twig/) (compare against this)

### Functional Review Steps
For each step, compare the PR-environment against the production link side-by-side — they should look and behave the same. Nothing here should look "new," just working again after the tooling change.

- [x] Sidebar navigation loads and is in the same order as production ([PR](https://deploy-preview-672--dev-component-library-twig.netlify.app/) / [production](https://yalesites-org.github.io/component-library-twig/)
  - [x] Confirm both show: Introduction, Tokens, Atoms, Molecules, Organisms, Templates, Page Examples, in that order
  - [x] Under Molecules, confirm "Cards" and "Modal" sit alphabetically (between Callout/Embed and Meta/Page Title) rather than jumping to the top
- [x] A component using icons/links renders correctly ([PR](https://deploy-preview-672--dev-component-library-twig.netlify.app/?path=/docs/molecules-site-alert--overview) / [production](https://yalesites-org.github.io/component-library-twig/?path=/docs/molecules-site-alert--overview))
  - [x] Verify the alert renders with its icon, heading, and a working dismiss button with
- [ ] A card with an external link renders its link-type icon ([PR](https://deploy-preview-672--dev-component-library-twig.netlify.app/?path=/story/molecules-cards--post-card) / [production](https://yalesites-org.github.io/component-library-twig/?path=/story/molecules-cards--post-card))
  - [ ] The headline link shows a small arrow icon (external link)
- [ ] A Docs (MDX) page renders normally ([PR](https://deploy-preview-672--dev-component-library-twig.netlify.app/?path=/docs/atoms-audio-player--overview) / [production](https://yalesites-org.github.io/component-library-twig/?path=/docs/atoms-audio-player--overview))
  - [ ] Intro text, the interactive preview, and the props table all appear
- [ ] A component's Controls panel live-updates the preview ([PR](https://deploy-preview-672--dev-component-library-twig.netlify.app/?path=/docs/molecules-site-alert--overview) / [production](https://yalesites-org.github.io/component-library-twig/?path=/docs/molecules-site-alert--overview))
  - [ ] On the Site Alert Docs page, find the "Alert Type" control and change it from `announcement` to `emergency`
  - [ ] The alert should immediately turn red with a warning-triangle icon (it starts out yellow with an info icon)
- [ ] Toolbar controls match production and visibly affect the preview ([PR](https://deploy-preview-672--dev-component-library-twig.netlify.app/?path=/docs/tokens-colors-theme-sandbox--docs) / [production](https://yalesites-org.github.io/component-library-twig/?path=/docs/tokens-colors-theme-sandbox--docs) — the "Theme Sandbox" page)
  - [ ] The theme dropdown reads "Site: Global Theme (lever)" (not just a color name like "Old Blues")
  - [ ] Switch it from "Old Blues" to "New Haven Green". The nav link accent colors and image overlays should shift from Yale blue to a darker green/black tone
  - [ ] Switch the typography dropdown from "YaleNew (Old-Style Numerals)" to "Mallory". The plain headings on the page (e.g. "Component themes (dials)") should switch from a serif to a sans-serif font. (The "Yale University" site-name/logo text intentionally does *not* change as that's a fixed brand element)
- [ ] **Visreg toggle still works** ([PR link](https://deploy-preview-672--dev-component-library-twig.netlify.app/?path=/docs/molecules-accordion--overview)). It switches to "Visreg: On" and reveals additional "Visreg" entries in the sidebar; click again to hide them

### Design Review
- N/A. There are no changes to the frontend. Any regressions should be considered a problem with this work.

### Accessibility Review
- N/A. There are no changes to the frontend. Any regressions should be considered a problem with this work.


miketullo95 (approved):
Approving — I checked out the Storybook multidev myself and confirmed it renders and behaves the same as production (sidebar order, icon/link rendering, MDX docs, live Controls updates, theme toggle, and the visreg toggle all working).

Two non-blocking notes for @dblanken-yale since they're more dev-side calls than mine:

1. **Atomic build check** — issue #1435's acceptance criteria calls out confirming `atomic`'s bundled dependency on `component-library-twig` still builds cleanly. I don't see that covered anywhere in this PR's testing checklist, and the build script changed meaningfully (webpack's `manifest.json`/`icons.svg` sprite generation is gone, Vite's `dist/` output is structured differently). Worth a quick check that `atomic`'s Drupal libraries still resolve against the new output, before or shortly after this merges.

2. **Node version bump** — this now requires Node >=24.13.0 org-wide (`package.json` engines, `.nvmrc`, all three GitHub workflows). That's a bigger footprint than just CI since every contributor working in this repo needs to update their local Node. Flagging in case that hasn't been communicated to the team yet.

@nJim nice work on this one — the inline comments explaining the trickier workarounds (sidebar sort, Twig function registration, Drupal shim timing) make this a lot easier for the next person to maintain.

miketullo95:
Tested the Netlify deploy against production, looks solid overall. Build works, and atoms/molecules/organisms/page examples all pass the eye test.

Two minor things worth a look, neither blocking:

1. Theme and typography toolbar controls (`config/emulsify-core/storybook/preview-decorators.js`) seem to trigger a hard refresh when changed. If I set the color dial to 2, then switch typography, the color dial resets back to default and I have to set it again. It also feels a bit slower than production's instant update. The `globalTypes`/decorator code itself looks unchanged from the old `.storybook/preview.js`, so this might just be how Storybook 10's Vite builder handles global state rather than something introduced in this migration. Curious if that's expected/known behavior on your end.

2. The external link icon on card headlines (`/story/molecules-cards--post-card`) renders, but looks very slightly smaller/thinner than production. Small enough that I almost missed it.

Not approving yet since there's a chance both of these turn out to need no action from you. Once you've had a chance to weigh in, I'll do a final pass and approve. @nJim

dblanken-yale:
Small leftover spotted while checking yalesites-org/YaleSites-Internal#1322 (Retire Percy) against this branch: the Percy removal here is complete in code — `package.json` and `package-lock.json` are both clean, `.percyrc` and `.github/workflows/visual-regression.yml` are deleted, and `README.md` is rewritten — but four prose mentions of Percy survive and now describe tooling this PR removes:

- `STORYBOOK-MIGRATION-GUIDE.md:150` — "Comprehensive coverage for Percy snapshots"
- `STORYBOOK-MIGRATION-GUIDE.md:242` — "These are static examples captured by Percy for automated visual testing."
- `STORYBOOK-MIGRATION-GUIDE.md:351` — "**Visual Testing**: Percy snapshots catch regressions automatically"
- `components/00-introduction/welcome.mdx:118` — "...for Percy visual regression testing coverage."

Worth folding into this PR rather than a follow-up, since they were orphaned by this diff. The README line 108 wording ("`[component].visreg.stories.js` pages remain as a documentation pattern ... independent of any visual-diffing tool") is a good template for the replacement text.

nJim:
@dblanken-yale @miketullo95 Sendign this back to you for review. There are two issues with this work that are not obvious to fix: the arranging of the side navigation and the loading of some component JavsScript. For both of these issues, I've contributed some Claude-generated code. These changes initially felt like overkill (especially the use of 'once' is many components) but I'm tempted to accept it at face value. In the future I may ask Callin from Four Kitchens to see if any of this can be pulled into Emulisify to simplify our component handling. I've taken this about as far as my skills can go. If you like, I can try to bring in a frontend engineer to take a lap if needed.

dblanken-yale:
Helped get this branch current with its base and gave the diff a review pass. Pushed six commits; below is what changed and, just as importantly, what I left alone for you.

### Reconcile

`after-2260` was behind `develop` in this repo and in yalesites-project, so per Dave those were fast-forwarded to `develop` first (both were strict ancestors — no merge commits, no rewritten history), then merged into this branch. All three branches now have their base as an ancestor.

The merge had three modify/delete conflicts, all resolved by keeping this branch's deletion — `webpack/plugins.js`, `.storybook/main.js`, and `.github/workflows/visual-regression.yml` (Percy retired by #694, out of scope per the ticket).

**One break git merged without a conflict, worth knowing about.** The base added committed placeholder PNGs at `images/placeholders/`; this branch had renamed `images/` → `assets/`. Git saw new files in a directory this branch no longer uses and merged them cleanly — leaving 34 fixture URLs pointing at `/images/placeholders/...`, which nothing serves (Emulsify Core mounts `assets/` at `/assets` and nothing at a bare `/images/`). A missing placeholder collapses a story's rendered height instead of erroring, so it would have shipped as a silent visual diff. Moved to `assets/images/placeholders/` and repointed the URLs, matching how you'd already migrated `assets/images/patterns/wavy.png`.

### Fixed, with the evidence

**Sidebar order didn't match production.** The comparator alphabetised stories *within* a component; production shows them in declaration order. 464 of 628 positions differed — theme stories read AI first instead of Old Blues, Standard Pages ended `...With Quick Links, With Sidebar` instead of `...With Sidebar, With Quick Links`. Also, Storybook 9's indexer emits a component's docs entry last, which sank "Overview" to the bottom of Molecules/Cards.

Worth flagging the method: comparing `.out/index.json` is the wrong layer, because production applies `storySort` client-side in the manager, so its index order isn't its sidebar order. I drove both builds in a browser, expanded every node, and read the tree in DOM order:

| Visreg | production | this branch |
|---|---|---|
| off | 138 items | 138 — **identical order** |
| on | 769 items | 769 — **identical order** |

**`npm ci` in atomic would have failed.** `files` is `["dist","components"]`, so the published tarball has none of `scripts/` (`npm pack`: 1137 files, 0 under `scripts/`), but `postinstall` ran three `node scripts/patch-*.mjs`. Running the old command inside the extracted tarball exits 1 with `Cannot find module .../patch-emulsify-core-twig-functions.mjs`. atomic's lock already records `hasInstallScript: true` for this package and its release workflow runs a bare `npm ci`. Guarded on `scripts/` existing; verified exit 0 both in the tarball and in the source checkout with `patch-package` still applying.

**`dist/tokens.json` stopped being emitted.** The deleted `webpack/plugins.js` copied it from the tokens package. `ys_themes/src/ColorTokenResolver.php` reads that exact path from PHP; on a miss it logs a warning and `getGlobalThemeColors()` returns `[]` — emptying Site Global Theme colour data site-wide with no build error. `copy-dist-fonts.mjs` is now `copy-dist-static.mjs` and carries both copies.

Also: hardened `patch-emulsify-core-twig-functions.mjs` (its guard compared the fully-replaced string, so a half-applied patch — imports added, functions never registered — exited 0); dropped `eager: true` from two icon globs that pulled all 97 SVGs into the preview bundle to read filenames; deduped the `autodocs` tag.

### Tests added

`npm run test` is green (95 unit tests, up from 87; lint, stylelint, prettier, Storybook build, plus a new `test:dist` after the build).

- `story-sort.test.mjs` — pins declaration order, docs-first, and that an explicit `__order` still wins. Verified the two regression cases fail before the fix and pass after.
- `fixture-asset-urls.test.mjs` — resolves every root-absolute image URL through the static mount table. This is the check that would have caught the placeholder merge.
- `dist-drupal-contract.test.mjs` — asserts every `dist/` file Drupal reads exists, each mapped to the consumer that breaks without it.
- `dist-module-format.test.mjs` — diffs the ES-module set under `dist/js/` against an allowlist, so a future code-split can't silently inert another Drupal behaviour.

### Left for you — not fixed on purpose

- **Asset versioning.** `getAssetPath()` is now a guaranteed passthrough (the Vite build emits no manifest). Narrower than it looks: everything in `atomic.libraries.yml` still gets Drupal's `?<system.css_js_query_string>`. The exception is `icons.svg`, referenced from `_yds-icon.twig` as a raw `<use xlink:href>` Drupal never rewrites — so that URL is now permanently stable, on essentially every page. After a release that changes an icon, a stale sprite renders new icons as empty space (a `<use>` at a missing `#id` fails silently). A cheap fix that needs no build support is to append the existing `system.css_js_query_string` state value. Left alone because it changes runtime output on every page and it's your call. Two related notes: the docblock says restoring hashing "is tracked separately" but cites no issue, and two unit tests are now permanently skipped everywhere, leaving that branch at zero coverage while reading as covered.
- **`components/00-introduction/introduction.stories.js` and `themes.stories.js` are now unindexed dead code** (~500 lines duplicating the MDX). `project.emulsify.json` declares structure roots for `00-tokens`..`05-page-examples` but not `00-introduction`, so core's `.stories.js` glob never scans it; the MDX survives only via your `components/[0-9]*/**/*.mdx` glob. No sidebar impact — both are `tags: ['!dev']` and never appeared — so this is a deletion call on your content, not a correctness fix.
- **This repo's `CLAUDE.md` is stale throughout** (node 20.14.0, `npm run develop` as webpack watch, `visreg:ci`, `.storybook/*` config, `manifest.json`/`tokens.json` as outputs). Didn't touch it since it's guidance you maintain.
- Lower priority: `get-url-type-function.js`/`get-asset-path-function.js` duplicate the bodies in `twig-url.js`/`twig-asset-path.js` and can drift; the behaviour-dedupe patch marks `<html>` for `context === document`, which is worth a manual pass over Docs pages and control changes (media-grid modal, tabs, accordion, utility nav).

dblanken-yale:
One more difference worth a decision, found while verifying the deploy preview against production.

**Production serves `/tokens.json`; the preview 404s on it.**

```
https://yalesites-org.github.io/component-library-twig/tokens.json     -> 200
https://deploy-preview-672--dev-component-library-twig.netlify.app/tokens.json -> 404
```

Cause: webpack copied `tokens.json` to `dist/tokens.json` and Storybook mounted `dist` at `/`. Emulsify Core mounts `dist/assets` at `/`, not `dist`, so the file is no longer reachable over HTTP even though the build now emits it again for Drupal.

Nothing inside this repo is affected — every component imports it at build time (`import tokens from '@yalesites-org/tokens/build/json/tokens.json'`), so there's no runtime fetch, and the sidebar/page parity checks all pass. I left it as-is rather than guess, because the only thing it could break is something *outside* this repo treating that URL as a public token feed. If you know of a consumer, the fix is a second copy into `dist/assets/tokens.json` (that mount makes it resolve at `/tokens.json` again); if not, this is fine to close out as an intentional change.

For completeness, the deploy preview on `335b8ada` was compared against production the same way as the local build — sidebar expanded fully and read in DOM order — and matched exactly: **138 items with Visreg off, 769 with Visreg on**, both identical order. `/assets/icons.svg` and `/assets/images/placeholders/placeholder-3x2.png` both return 200 on the preview.

dblanken-yale:
Ran a `/simplify` pass over what I'd changed and pushed four more commits. Two of them are real bugs, one of which is in local dev rather than CI — worth reading.

### Watch mode was wiping `dist/fonts` and `dist/tokens.json`

Vite clears its prepare-out-dir state in the `options()` hook, which runs at the start of **every** build in a watch session, so `renderStart` re-empties `outDir` on each rebuild. `npm run vite` and `npm run develop` never invoke the copy step chained after `vite build`, so nothing puts them back.

Measured here: both files present, then both **MISSING** after the first watch build. For anyone running `npm run develop` against a local Drupal site that means no fontawesome (atomic reads `dist/fonts/fontawesome/...`) and no Site Global Theme colour data (`ys_themes` reads `dist/tokens.json`) for the whole session, with no error explaining it. The fonts half predates this branch — I'd just restored `tokens.json` into the same mechanism.

Both copies now live in a `writeBundle` plugin with `enforce: 'post'`, the same hook and ordering Core uses for its own `emulsify-copy-all-src-assets`, so they run per build and watch is covered. Verified they survive the first watch build and a subsequent rebuild.

### A false positive in the guard I added

`fixture-asset-urls.test.mjs` listed only `assets/ → /assets`, but Core also mounts `dist/assets` at `/assets` and at `/` — and `icons.svg` exists *only* in `dist/assets`, since the sprite is generated from `assets/icons/*.svg`. So a fixture referencing `/assets/icons.svg` was reported unserved even though Storybook serves it fine. I confirmed it by adding one and watching the guard fail.

Fixed with a named build-generated exception rather than deriving the table from `buildAssetStaticDirs()` — deriving would accept anything present in a built `dist/`, which makes the guard strictly weaker.

### Housekeeping

Shared the byte-identical file walk between the two fixture guards (the duplicated part is a *policy* — which file types can carry an image URL — and drift there is silent); shared one `patchFile` across the emulsify-core patch scripts, since the per-anchor verification I'd added already existed correctly in the sibling and two copies is exactly how the first came to be missing it; made the dist guards fail loudly instead of skipping (standalone `npm run test:dist` was reporting green while checking nothing); and replaced the `ORDER_TREE` copy in the Storybook guide with a shape-only sketch, since the pasted copy had already drifted from the real tree.

### Worth a decision: the placeholder PNGs are 42.3 MB

The seven files are 4.1–7.0 MB each. `assets/images/placeholders/README.md` says "116 KB for all seven" and gives an ImageMagick flat-fill recipe that cannot produce multi-megabyte PNGs, so the committed bytes are not what the documented command generates.

**This is inherited, not from this PR** — the blob SHAs are identical to the base branch (`8c8b33ed…`, added by `2de29e35 fix(1611): self-host visreg placeholder images`). Flagging it because my commit is the one that moved them, and 45 MB of them get copied into `.out`, so every deploy preview and the gh-pages build carries them and the image stories fetch multi-megabyte files. They're **not** in the npm tarball (`files` is `["dist","components"]`), so atomic is unaffected. Re-running the documented recipe would likely shrink them by ~99%, but that changes the images, so it's your call.

### Deliberately not changed

- The `orderList` guard in the same-title sort branch measures as having zero effect on today's index, but I kept it: it makes `ORDER_TREE`'s explicit intent authoritative instead of depending on the indexer's emission order, and an implicit dependency on emission order is what caused the ordering regression in the first place.
- Six efficiency findings, all measured and negligible (the duplicate tree read is ~20 ms of a 720 ms suite, and a shared module wouldn't remove it since `node --test` forks per file; the sort comparator is 0.632 ms total for 758 entries).

### Final state

`npm run test` exits 0 (95 unit tests, lint, stylelint, prettier, Storybook build, both dist guards). The Netlify preview on `541d5465` is green and I re-verified it against production: **138 items identical with Visreg off, 769 identical with Visreg on**. I also re-ran the yalesites-project Pantheon deploy so the multidev builds against these updated companions — **success**.

Cleared so nobody re-litigates it: the `micromodal.es-*.js` chunk the calendar imports **does** ship in the published tarball and resolves correctly from the calendar's relative import (verified inside the extracted tarball).

dblanken-yale:
Second code-review pass over my own fixes turned up three more, now pushed (`a0d8bfee`). One of them was a hole in a guard I'd added.

### The module-format guard could miss the exact failure it exists to catch

The regex anchored on a line start, but Vite output is minified to two lines and Rollup puts `export{…}` at the **end** of the code line. So it missed `a=1;export{b};`, `function f(){}export default f;`, and an import behind a banner comment — all in the dangerous direction.

The realistic path to that: component A imports a helper from component B's *entry* module, so Rollup keeps B's exports. B becomes an ES module with a trailing `export{…}`, the guard reports the set unchanged, atomic never gets `type: module`, and B's behaviour goes inert with CI green — precisely the silent failure the guard is for.

Replaced the regex with `new vm.Script(code)` in a try/catch. That asks the question the browser actually asks — *does this compile as a classic script* — instead of pattern-matching a proxy for it. `node:vm` is a built-in, so no new dependency, and it's 10 ms for all 33 files. Verified it's correct on all six shapes above (and still ignores `import(`, `import.meta`, and the string literals `"import"`/`"export"` inside highlight.js's keyword arrays), flags exactly modal + calendar on the real `dist`, and catches a planted `var a=1;export{a as b};` that the regex passed.

### `String.replace` was eating `$` patterns in injected code

`patch-emulsify-core-storybook-story-sort.mjs` inlines the **entire text** of `story-sort.mjs` into Core's `preview.js` as a replacement string, and `String.replace` interprets `$&` / `$1` there. A `$1` anywhere in that file would be silently substituted, corrupting the inlined comparator — best case a build parse error, worst case a subtly wrong sidebar order visible only in the built output. There's no `$` in it today, and that script already uses `'$1'` in a replacement of its own, so it's one regex helper away.

Switched both patchers to replacer functions, which are never scanned for `$` patterns — `$`-proof by construction rather than by audit. Verified from a truly pristine `@emulsify/core` (`npm ci --ignore-scripts`, so lifecycle scripts didn't pre-patch it): all three patchers apply, `storySort` and `ORDER_TREE` land, comparator textually intact.

Also folded `npx patch-package` inside the `scripts/` guard so the whole postinstall is a silent no-op for consumers (verified against the packed tarball: exit 0, zero output), and fixed a pointer to the since-deleted `copy-dist-static.mjs`.

---

## Separate issue, inherited from the 1611 placeholder work — worth raising with its author

While verifying, I measured the committed placeholder PNGs. **Six of the seven have the wrong aspect ratio**, and all seven are ~1000x larger than documented:

| file | actual | README says | |
|---|---|---|---|
| `placeholder-1x1.png` | 2250x1500 (1.5) | 2400x2400 (1.0) | mismatch, 7.0 MB |
| `placeholder-4x3.png` | 2249x1500 (1.50) | 2400x1800 (1.33) | mismatch, 6.6 MB |
| `placeholder-3x2.png` | 2260x1499 (1.51) | 2400x1600 (1.5) | ok, 7.0 MB |
| `placeholder-16x9.png` | 2250x1500 (1.5) | 2400x1350 (1.78) | mismatch, 4.9 MB |
| `placeholder-2x3.png` | 2264x1500 (1.51) | 1600x2400 (0.67) | **landscape, should be portrait**, 6.5 MB |
| `placeholder-5x8.png` | 2250x1500 (1.5) | 2400x3840 (0.63) | **landscape, should be portrait**, 4.1 MB |
| `placeholder-6x1.png` | 2000x1500 (1.33) | 1200x200 (6.0) | **should be a wide banner**, 6.3 MB |

They're all essentially the same 3:2 landscape image, and they're photographs — not the flat-fill placeholders the README documents with an ImageMagick recipe claiming "116 KB for all seven." Actual total: **42.3 MB**.

This matters because the README states the reason the ratios were chosen: `_image.twig` emits no `width`/`height`, so the placeholder's real ratio drives rendered height, and matching ratios "leaves measured story heights unchanged." That premise doesn't hold as shipped — so `image__alt: 'A 2 by 3 image'` renders a landscape image, and every ratio-keyed fixture renders at the wrong shape. Each `srcset` also lists 14 width descriptors all pointing at the same multi-megabyte file, and 45 MB of these get copied into `.out`, so every deploy preview and the gh-pages build carries them.

**This is not from this PR** — the blob SHAs are identical to the base branch (`8c8b33ed…`, added by `2de29e35` / `aa06041a`). My merge is only what repointed the fixture URLs at them. Flagging rather than changing it, since regenerating the images is the 1611 author's call.

To be precise about what I did and didn't verify: my parity checks cover sidebar structure and ordering, page rendering, and asset resolution — **138 items identical with Visreg off, 769 identical with Visreg on** against production, re-confirmed on the deployed preview for `a0d8bfee`. They do not compare rendered image pixels, and image stories will legitimately differ from the published build because the base replaced `picsum.photos` with these committed files. That's an intentional separate change, not migration drift.

`npm run test` exits 0 (95 unit tests, lint, stylelint, prettier, Storybook build, both dist guards). The yalesites-project Pantheon deploy was re-run against this companion state — green.

dblanken-yale:
Dave walked the component library preview and found five things. All five are addressed in `27b3e8ba`; for each one I checked the **same page on the published production build first**, so "we broke it" and "it was always broken" are separated by evidence rather than assumption.

| # | report | production | this branch | verdict |
|---|---|---|---|---|
| 1 | "Reset dismissed alerts" does nothing | works | **broken** | our regression |
| 2 | Read time not visible | hidden too | hidden | pre-existing |
| 3 | Video Background shows no video | no video | no video | pre-existing (dead URL) |
| 4 | Cannot traverse calendar months | doesn't change | doesn't change | pre-existing; needs Drupal |
| 5 | Menu docs pages need height | — | — | enhancement |

### 1. Reset dismissed alerts — our regression

The stories defined `resetAlerts` in an inline `<script>` inside the returned markup, and Storybook inserts a story's HTML with `innerHTML` — where per spec a `script` element **never executes**. Production: the click clears the keys, console shows only a harmless `Identifier 'resetAlerts' has already been declared`. Here: keys untouched, `resetAlerts is not defined`.

Moved into a module the stories import for its side effect. Verified on the static build: dismiss writes `ys-alert-id-123`, reset clears it to `[]`, all four alerts restored, **zero console errors**.

One trap worth recording: `typeof window.resetAlerts` is `undefined` *even when this works*, because `const` at a classic script's top level creates a global lexical binding, not a `window` property. The click's effect is the real test, not the global.

### 2. Read time — pre-existing

`yds-read-time.scss` hides the component and only reveals it under `html.js` or `html:has(.yds-storybook-cl)` — its value is computed at runtime, so there's nothing to show without JS. Drupal adds `js` to `<html>`; Storybook has no Drupal bootstrap, and **nothing in the repo ever sets `yds-storybook-cl`**. Production computes to `display: none` too. The text was correct all along — "Estimated read time: 3 min" was sitting in the DOM at 0×0.

Added the `js` class in `preview-head.html`, the same place and timing Drupal does. Chose that over `yds-storybook-cl` on measured blast radius: `html.js` appears in exactly **one** rule in the whole library, while `yds-storybook-cl` also drives `.yds-layout { margin-block: 0 }`, which would shift layout spacing across many stories.

### 3. Video Background — pre-existing, now self-hosted

Dave asked whether the mp4 is gone. More specific than that: the archive.org **item still returns 200 and still lists the file**, but the URL hardcodes a *specific* storage node and archive.org has moved the item — its own metadata now names `dn711107.ca` / `dn600306.us` as workable, not `ia800301.us`, which times out (HTTP 000 after 20s). Repointing at today's node would rot the same way.

So it's committed instead, same reasoning that moved the placeholder images off `picsum.photos`: `assets/videos/placeholder-loop.mp4`, 1280×720, 10s, H.264, **75 KB** — kept deliberately tiny. ffmpeg recipe and licensing in a README beside it. Verified: `readyState: 4`, 1280×720, `currentTime` advancing.

### 4. Calendar months — needs Drupal, and **works there**

`yds-calendar.js` calls `Drupal.ajax()` to fetch a month, and neither the old nor the new Storybook shim provides `Drupal.ajax` — production's month label doesn't change either. I deliberately did **not** shim it: a no-op `Drupal.ajax` would fail silently, which is worse than an error.

Verified on Lando instead, and this is the payoff for atomic #495's `type: module` fix: `Drupal.behaviors.eventsCalendar` **is registered** (before that fix the file was a parse-time SyntaxError and never ran), `Drupal.ajax` is a function, and the month advanced **July 2026 → August 2026** on click.

### 5. Menu docs pages — done

Diagnosed rather than assumed: nothing was clipping the dropdowns — every ancestor is `overflow: visible`. They're `position: absolute`, so they add no height, leaving a ~65px canvas with a ~520px menu drawing over the docs below.

Added `.sb-tall-canvas` beside the existing `.sb-contained-canvas` helper, with heights measured per page: primary nav and content collection at 38rem (~520px and ~375px dropdowns), utility nav at a `--short` 19rem since its dropdown is only ~137px under a ~44px bar. Breadcrumbs left alone per Dave. Verified with the menu open on each: 608 vs 520, 608 vs 375, 304 vs 161 — all inside the canvas.

(Aside: the `secondary-nav/` directory actually powers the **Content Collection** page — `title: 'Organisms/Menu/Content Collection'`.)

---

## Also verified in Lando Drupal, per Dave's request

**The local environment was broken first, and would have invalidated any result.** `atomic/node_modules/@yalesites-org/component-library-twig` symlinked *out* of the checkout to a different clone; only `/app` is mounted in the container, so the local Drupal site could load **no** component-library CSS or JS at all. Running `gyst` fixed it properly (and produced the same relative target), after which `style.css`, `yds-calendar.js` and the `micromodal.es-*.js` chunk all serve 200 in-container.

No new content was created — the site carries a "DO NOT CHANGE CONTENT" visual-regression warning and existing nodes already covered every case.

- **Calendar:** month traversal works (above).
- **Read time:** works, and always did — `<html class="no-touchevents js">`, `display: block`, "Estimated read time: 3 min". Confirms the Storybook fix targeted the real gap rather than hiding a component bug.
- **Site alert:** behaviour registered; toggling went `expanded → collapsed`, wrote `ys-alert-id-1764172791`, and survived a full reload. The reset *button* is a Storybook-only story affordance, so there's nothing to reproduce in Drupal and I'm not claiming otherwise.
- **Video background:** node 70 renders no `<video>` at all, because `atomic/templates/field/field--media--field-media-video-file--background-video.html.twig` embeds the component with `video_background__content: item.content` — the uploaded media file. So the twig `default(...)` I repointed never fires on a configured Drupal video. No Drupal regression, and no Drupal fix claimed.

`npm run test` exits 0 — 95 unit tests, eslint, stylelint, prettier, Storybook build, both dist guards. Screenshots (before/after on the deployed preview, on the static build, and in Drupal) are saved locally and can be attached if useful.

dblanken-yale:
Two follow-ups pushed: the Storybook bump Dave asked about, and a real `npm run develop` failure it surfaced.

## Storybook 10.5.4 → 10.5.10 (`4196e723`)

Worth doing, and the reason isn't "it's newer": the declared range was **already** `^10.1.4` (emulsify-core declares the same, with no hard pin), so the lockfile was the only thing holding 10.5.4. `npm outdated` reported `wanted: 10.5.10` — meaning a fresh `npm install` resolved 10.5.10 while `npm ci` resolved 10.5.4, so contributors and CI had quietly diverged. Pinning the floor to the version we actually test closes that.

Patch-only, and two entries are directly ours:

- **10.5.9** — "Preview: Fix crash when initialising UrlStore on a docs path." Every component's Overview page *is* a docs path.
- **10.5.10** — docs tooltips no longer inherit the wrong font. This project loads the Yale webfonts.
- Security: `ws` advisories (10.5.5) and a Vitest CVE (10.5.10). Both dev-only, neither ships in `files: ["dist","components"]`, so no consumer exposure — but the repo already carries a large Dependabot backlog.

Everything else in the range is TanStack/Angular/Vue/ReactNative/Webpack — unused here.

**Checked the things a bump can quietly break:** `--save-dev` can move a dependency between sections, and it did not — all five were already in `devDependencies`, `@emulsify/core` is still a production dep and storybook is still dev-only (this matters because yalesites-project builds with `npm ci --omit=dev`). All three postinstall patches still apply against the new tree. Gate exits 0 building storybook v10.5.10. Sidebar still byte-identical to production: **138 items Visreg off, 769 on**. All four of the earlier fixes re-verified on it.

### Left for a deliberate decision: `@emulsify/core` 4.2.1 → 4.4.0

`npm outdated` also reports this, and I deliberately did **not** take it. It's a *minor* bump of the package this whole migration sits on, and the three postinstall scripts patch its internals by exact string anchors — the hardened patcher would now fail loudly rather than half-apply, which is the right signal, but it needs its own verification cycle and could shift build output or sidebar order. It's arguably inside the ticket's "update Emulsify to its current version" AC, so it's your call whether it belongs here or in a follow-up.

## `npm run develop` was broken (`ed7b13c6`)

Dave reported it dying with:

```
Error: Failed to load static files, no such directory: ./dist/assets
Broken build, fix the error above.
```

...alongside "Importing a module script failed" when visiting Utility Nav. **Those were one bug.** Once the dev server exits on the broken build it stops serving module chunks, so whatever story is open reports the module error. That's also why the module error wouldn't reproduce on its own — I loaded Utility Nav directly *and* via sidebar navigation, on both the dev server and the deployed preview, and it rendered every time with no module errors.

Cause: `develop` runs the vite watch build and the Storybook dev server concurrently. Vite empties `outDir` at the start of **every** build in a watch session, and Storybook's `staticDirs` include `dist/assets`. So a developer who already has a build hits a race — Storybook resolves staticDirs while `dist/assets` exists, vite empties `dist`, and Storybook's favicon lookup calls `parseStaticDir` on a directory that just vanished. On a *cold* start it doesn't happen (Storybook gets there first, the dir is absent, and `existingStaticDirs()` filters it out), which is why this only bites normal day-to-day use.

Fixed at the cause: `vite` now passes `--emptyOutDir=false` — emptying the output directory on every rebuild is destructive and pointless in a watch loop, and it's the same behaviour that was deleting `dist/fonts` and `dist/tokens.json` mid-session. `ensure-dist` also creates `dist/assets` so the directory exists from the first moment. `npm run build` is untouched and still clean-builds.

Verified in the configuration that actually fails (warm `dist`): zero errors, Storybook reaches "ready", dev server answers 200, and `dist/assets` / `dist/fonts` / `dist/tokens.json` all survive the watch build — previously the latter two were destroyed.

The "Generated an empty chunk" warnings seen alongside are pre-existing and unrelated: three files under `00-tokens`, `04-page-layouts` and `05-page-examples` export no runtime code.

dblanken-yale:
Attempted `@emulsify/core` 4.2.1 → 4.4.0. **Patches were the one thing that wasn't a problem** — but there's a Drupal blocker, plus two genuine issues found along the way (one now fixed and pushed in `1bf07ce2`).

## Patches: zero new ones needed

I packed and diffed both versions before touching the repo. Every file the three postinstall scripts anchor into is **byte-identical** between 4.2.1 and 4.4.0 — `function-map.js`, `platform-behaviors.js`, `_drupal.js`, `preview.js`. Confirmed empirically after bumping: postinstall exits 0 and every marker lands. I also checked whether 4.4.0 added an extension point that would let us *delete* the twig-functions patch — it hasn't.

Wider surface: 28 files differ, plugins reorganised into `plugins/{assets,reporter,twig}`. `buildAssetStaticDirs` was refactored onto a shared `resolveAssetRoots()` but the resulting mounts are functionally identical, so the mount table in `fixture-asset-urls.test.mjs` stays accurate.

## Blocker 1: rolldown filename length — fixable, and it's gyst-specific

4.4.0 requires Vite 8 (so does **all** of 4.3.x — there's no intermediate Vite 7 release). Vite 8 bundles with rolldown, which names chunks after the module id, and Emulsify's virtual Twig modules embed the *full absolute source path* with `/` escaped as `_2F` — 3 bytes per slash. The Storybook build died:

```
Failed to write file in ./.out/storybook-assets/_virtual_emulsify-twig-dep__2FUsers_2F...
Caused by: File name too long (os error 63)
```

Measured: **257 bytes against macOS's 255 limit — over by 2.**

To be precise, since it came up: **gyst isn't broken.** gyst's *layout* nests this package at `.../atomic/_yale-packages/component-library-twig` (99 chars) versus ~41 for a standalone clone and ~62 in CI. Because the filename grows ~3× faster than the path, that depth is what crosses the limit. The failure mode is worse than a clean break: **CI and Netlify build green while every gyst'd developer machine fails on the identical commit.**

Fixable without a patch — an `output.sanitizeFileName` override (truncate at 180, replace the tail with a stable hash) makes the build succeed. Verified, but not committed: it belongs with the upgrade.

## Blocker 2 (the real one): rolldown's shared runtime breaks Drupal JS aggregation

`scripts/dist-module-format.test.mjs` — the guard added earlier in this PR — failed on the first run:

```
actual:   yds-text.js, yds-modal.js, yds-calendar.js, link-treatment.js
expected: yds-modal.js, yds-calendar.js
```

rolldown emits a shared `rolldown-runtime-*.js` chunk, so two more Drupal-consumed files become ES modules:

```
dist/js/components/01-atoms/typography/text/yds-text.js  → import{...}from"../../../../../assets/rolldown-runtime-*.js"
dist/js/link-treatment.js                                → import{...}from"../assets/rolldown-runtime-*.js"
```

atomic loads both as classic scripts, so both would be parse-time SyntaxErrors with their behaviours never registering. Fixing it means `type: module` + `preprocess: false` on those libraries — and **`link-treatment.js` is in atomic's `global` library, which loads on every page.** `preprocess: false` there **removes site-wide JS aggregation for every YaleSites page**: a real front-end performance regression traded for a Storybook-side gain. And because the ES-module set is now decided by rolldown's chunking heuristics, it can shift again on any future patch release.

Worth noting this is exactly the silent failure that guard exists for. Without it, 4.4.0 would have looked clean in Storybook and quietly inerted a site-wide behaviour in Drupal.

**Recommendation: don't land 4.4.0 here.** It wants its own change with Drupal QA, since the risk concentrates in Drupal. Everything needed is written up above.

## Fixed and pushed: `npm ci` was broken (`1bf07ce2`)

The investigation surfaced a genuine latent bug. `@emulsify/core` was declared `^4.0.0`, which lets npm resolve **4.4.0** while the lockfile pins 4.2.1. So `npm ci` — what CI and yalesites-project's frontend build both run — failed:

```
Missing: @emnapi/core@1.11.3 from lock file
Missing: @emnapi/runtime@1.9.2 from lock file
```

(`@emnapi/*` are rolldown's napi bindings, dragged in by the 4.4.0 resolve.) Pinned to `~4.2.1`: patch updates stay available, the 4.2 line is held, and the caret can no longer silently adopt a version we've now proven unsafe. `npm ci` exits 0, patches apply, `npm run test` exits 0 on storybook v10.5.10, sidebar still 138 / 769 identical to production.

## Also worth a decision: this PR has no CI test gate

While checking whether `Test` had run on my pushes, I found it never has on this branch. `.github/workflows/test.yml` triggers on `pull_request: branches: ['main','develop']`, but **#672 targets `after-2260`** — so the job is never scheduled. The only automated check here is the Netlify build.

Two consequences: local `npm run test` has been the sole gate for this migration, and the `npm ci` breakage above wouldn't have surfaced until the PR is retargeted to `develop` — at which point `Test` runs for the first time, starting with `npm ci`. I haven't touched the triggers (that's a repo-policy call), but it should be a conscious one.

dblanken-yale:
Emulsify Core 4.4.0 is now tracked separately in yalesites-org/YaleSites-Internal#1619 rather than riding on this PR — the blocker is that Vite 8's bundler makes `dist/js/link-treatment.js` a JS module, and that file is in Atomic's site-wide `global` library, so supporting it costs site-wide Drupal JS aggregation. The full investigation (including the working `sanitizeFileName` fix for the 257-byte filename failure, and confirmation that no new patches are needed) is written into that ticket. `@emulsify/core` stays pinned to `~4.2.1` here.

